PR #4, opened 2021-04-14
Add file_encoding support

6cdh:
Encoding of a file is common information. So I added it. Use

```
provider = 'file_encoding'
```


Famiu:
I made some minor modifications merged it. Thank you for your contribution.

PR #8, opened 2021-04-15
Experimental async support

famiu:
I have implemented experimental async support for Feline, it would be nice if people test it before I merge it

### DIFF
--- a/lua/feline/init.lua
+++ b/lua/feline/init.lua
@@ -1,5 +1,6 @@
-local g = vim.g
-local fn = vim.fn
+local wo = vim.wo
+local uv = vim.loop
+
 local gen = require('feline.generator')
 local utils = require('feline.utils')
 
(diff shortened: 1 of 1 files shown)


jchilders:
So.. I've been using feline for a couple of days, and really like it. I came here to report a problem, though: while normally feline is very fast, it slows down dramatically when the current buffer is attached to an LSP.

Is this PR to address that issue?

famiu:
> So.. I've been using feline for a couple of days, and really like it. I came here to report a problem, though: while normally feline is very fast, it slows down dramatically when the current buffer is attached to an LSP.
> 
> Is this PR to address that issue?

Using async features might make the statusline faster under certain conditions, yes. But at the end it depends. But this PR is not about that. 

And as for slowdown while an LSP client is attached, I never faced any slowdowns and I have a really low-end laptop with only 2GB of RAM. Maybe try running Neovim without other plugins to see if it's really Feline's fault. If that doesn't fix it, make a separate issue and I'll look further into it.

famiu:
> I have implemented experimental async support for Feline, it would be nice if people test it before I merge it

Just noticed the async branch doesn't update statusline contents after the statusline is once loaded. Have to look into it now

jchilders:
So looks like this branch fixed the performance issues I was seeing. Thank you.

famiu:
> So looks like this branch fixed the performance issues I was seeing. Thank you.

This branch seems to still have several issues though. For one, it doesn't update the statusline after loading it so it won't update cursor position or scrollbar. I'm working on fixing it though.

famiu:
So far I have no idea on how to fix the statusline not being updated. It seems to be because of the fact that the async function can't return and thus vim.wo is set to a fixed string value instead of a function. I don't know how to work around that. I might have to abort the async support if I can't figure out a solution

jchilders:
May want to look in to what galaxyline, airline, or other Lua-based statusline plugins are doing.

famiu:
> May want to look in to what galaxyline, airline, or other Lua-based statusline plugins are doing.

Lualine does it synchronously, I believe airline does too. Galaxyline is async and it seems to use a hardcoded list of autocmd events, and I don't want to just copypaste it without properly understand how it works to avoid pitfalls. Maybe it'd just be fine if Feline remained synchronous as well.

famiu:
I've managed to fix the async statusline not updating issue. Would still want to wait to see if async actually offers any significant performance improvements before merge.

famiu:
I've thought for a while and decided that I'm just going to abort the async support. It's just more complicated and a lot less smooth compared to the synchronous version.

Issue #21, opened 2021-07-18
[Feature] Allow setting hl from colorschemes via name


rktjmp:
_Great plugin._

I am trying to set my highlights via my colorscheme. Since feline only defines highlights on demand (afaict), any I define get clobbered.

I am using Lush, so it's actually not that big a deal to just import the theme data into feline and do it but given I can set the name, maybe you could check if the given name exists before defining it?

I can see a few ways to do this:

**Check via vim.fn.hlexists()**

```lua
-- Add highlight of component
local function add_component_highlight(name, fg, bg, style)
    local hlname = 'StatusComponent' .. name

    -- check via vim.fn.hlexists,
    -- may incur some viml bridge cost so perhaps checking 
    -- and memoising the result into highlights[hlname] so future 
    -- checks are speedy?
    if highlights[hlname] or (vim.fn.hlexists(hlname) == 1) then
       -- i.e highlights[hlname] = "externally_defined"
        return hlname
    else
        cmd(string.format('highlight %s gui=%s guifg=%s guibg=%s', hlname, style, fg, bg))
        highlights[hlname] = true
        return hlname
    end
end
```

**Allow for the `hl` option be a string or function that returns a string**

If `hl` is a string, just apply the highlight given

```lua
components.left.active[1] = {
    provider = 'file_type',
    hl = "MyFileTypeGroup"
}

components.left.active[1] = {
    provider = 'file_type',
    hl = function() return "MyFileTypeGroup" end
}
```

```lua
local function parse_hl(hl)
    if hl == {} then return defhl end
    -- maybe prefix with StatusComponent, but probably just leave as given
    -- since the user is being explicit about what they want and can namespace
    -- the group themselves if they want
    if type(hl) == "string" then return hl end
```

**Don't generate a highlight group if no data is given**

```lua
components.left.active[1] = {
    provider = 'file_type',
    hl = {
      -- this will generate the group, with the 
      -- name StatusComponentMyGroup fg = red
      fg = "red",
      name = "MyGroup"
    }
}

components.left.active[1] = {
    provider = 'file_type',
    hl = {
      -- this will not redefine the group
      name = "MyGroup"
    }
}
```

I can submit a PR if you have a preferred method, probably #2 is neatest?

(I like the name.)

rktjmp:
PR doesn't have docs, can write if you want to accept it.

PR #22, opened 2021-07-18
feat: Use hl as hlgroup if hl is a string or fn->string

rktjmp:
If a components hl option is a string, or a function that returns
a string, accept it as an existing or externally defined hlgroup
and don't generate our own.

See #21


rktjmp:
Also allows for `left|right_sep.hl` to be a function.

Issue #23, opened 2021-07-24
[Bug] feline causes the neovim splash screen to disappear


ibhagwan:
**Neovim version**
Both 0.6 nightly and 0.5 release:
```
NVIM v0.6.0-dev+63-ga8c3d50fa
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```
```
NVIM v0.5.0
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

**Feline version**
`4c04c967691d028b4056f726735db329039cdff2`

This was happening since the first version of `feline` I used, this isn't a new issue.

**Describe the bug**
Feline causes the neovim spash screen to disappear, for clarity I'm referring to the opening screen with the neovim version, etc:
```
~                     NVIM v0.5.0
~   
~      Nvim is open source and freely distributable
~                 https://neovim.io/#chat
~   
~     type  :help nvim<Enter>       if you are new!
~     type  :checkhealth<Enter>     to optimize Nvim
~     type  :q<Enter>               to exit
~     type  :help<Enter>            for help
~   
~              Help poor children in Uganda!
~     type  :help iccf<Enter>       for information
```

**To Reproduce**

Run feline as the only plugin and it should reproduce, it even happens when lazy loading to with packer on `VimEnter`:
```lua
use { 'famiu/feline.nvim',
  requires = { 'kyazdani42/nvim-web-devicons' },
  config = "require'plugin.feline'",
  event = 'VimEnter'
}
```

**Expected behavior**
Keep the splash screen.

**Does this error occur in the minimal init file?**

Yes, the issue is reproducible with `minimal_init.lua`.

**Additional context**

Most fancy status line plugins I tried have the same issue, for exmaple [galaxyline issue #107](https://github.com/glepnir/galaxyline.nvim/issues/107), `gitsigns` used to have a similar problem, perhaps [gitsigns issue #238](https://github.com/lewis6991/gitsigns.nvim/issues/238) will provide a clue how to fix it?

famiu:
As of now I have no idea what could be causing that to happen. I might look into it further when I'm not busy but in  the meantime, if anyone knows what could be causing it, it'd be great if they could inform me of it

ibhagwan:
> As of now I have no idea what could be causing that to happen. I might look into it further when I'm not busy but in the meantime, if anyone knows what could be causing it, it'd be great if they could inform me of it

[...]

PR #24, opened 2021-07-24
fix for issue#23, feline makes splash screen disappear

ibhagwan:
This closes [issue #23](https://github.com/famiu/feline.nvim/issues/23).

[This comment in Powerline issue #250](https://github.com/powerline/powerline/issues/250#issuecomment-52275269) gave me the lead with:
```vim
" Hack to show startup screen. Problems in GUI:
" - Defining local value of &statusline option while computing global value 
"   purges startup screen.
" - Defining highlight group while computing statusline purges startup 
"   screen.
" This hack removes the “while computing statusline” part: both things are 
" defined, but they are defined right now.
```

I figured if I just call `generate_statusline` twice (`true|false`) this will generate all required `highlight` groups so that the `highlight` command isn't called as part of the initial `vim.o.statusline` generation:

```lua
    gen.generate_statusline(true)
    gen.generate_statusline(false)
```

### DIFF
--- a/lua/feline/init.lua
+++ b/lua/feline/init.lua
@@ -76,6 +76,19 @@ function M.setup(config)
     generator.components = components
     generator.properties = properties
 
+    -- fix for issue #23, according to:
+    -- https://github.com/powerline/powerline/issues/250
+    --  + Defining local value of &statusline option while
+    --    computing global value purges startup screen.
+    --  + Defining highlight group while computing statusline
+    --    purges startup screen.
+    -- The below calls will make sure all highlight groups
+    -- are already created and saved in `highlights` table
+    -- located at generator.lua#L33 so feline doesn't run
+    -- `highlight` commands as part of status line genration
+    gen.generate_statusline(true)
+    gen.generate_statusline(false)
+
     vim.o.statusline = '%!v:lua.require\'feline\'.statusline()'
 
     create_augroup({


famiu:
> I figured if I just call `generate_statusline` twice (`true|false`) this will generate all required `highlight` groups so that the `highlight` command isn't called as part of the initial `vim.o.statusline` generation:
Incorrect. There are components (and highlights) that could be conditionally activated. In which case their highlights would still not be defined

ibhagwan:
> > I figured if I just call `generate_statusline` twice (`true|false`) this will generate all required `highlight` groups so that the `highlight` command isn't called as part of the initial `vim.o.statusline` generation:
> 
> Incorrect. There are components (and highlights) that could be conditionally activated. In which case their highlights would still not be defined

Obviously the best solution would be to have a “setup_highlights” method you can call before generating the status line which would create all possible highlights but that is out of the scope of this PR, I gave you everything you need to solve it properly if you wish.

famiu:
> > > I figured if I just call `generate_statusline` twice (`true|false`) this will generate all required `highlight` groups so that the `highlight` command isn't called as part of the initial `vim.o.statusline` generation:
> > 
> > 
> > Incorrect. There are components (and highlights) that could be conditionally activated. In which case their highlights would still not be defined
> 
> Obviously the best solution would be to have a “setup_highlights” method you can call before generating the status line which would create all possible highlights but that is out of the scope of this PR, I gave you everything you need to solve it properly if you wish.

What about the solution that creates highlights asynchronously? Does that work?

ibhagwan:
> > > > I figured if I just call `generate_statusline` twice (`true|false`) this will generate all required `highlight` groups so that the `highlight` command isn't called as part of the initial `vim.o.statusline` generation:
> > > 
> > > 
> > > Incorrect. There are components (and highlights) that could be conditionally activated. In which case their highlights would still not be defined
> > 
> > 
> > Obviously the best solution would be to have a “setup_highlights” method you can call before generating the status line which would create all possible highlights but that is out of the scope of this PR, I gave you everything you need to solve it properly if you wish.
> 
> What about the solution that creates highlights asynchronously? Does that work?

I’m not sure which solution you mean, I tried both `schedule_wrap` and async loop and they didn’t work.

I still believe this is the best and simplest solution, even if later down the life of the session there will be a dynamically generated highlight it doesn’t matter, all we’re trying to solve here is the intro screen, calling the generator twice seems to solve it as the other dynamically generated highlights will usually happen in a file buffer at which point you lost the intro screen anyways.

famiu:
> > > > > I figured if I just call `generate_statusline` twice (`true|false`) this will generate all required `highlight` groups so that the `highlight` command isn't called as part of the initial `vim.o.statusline` generation:
> > > > 
> > > > 
> > > > Incorrect. There are components (and highlights) that could be conditionally activated. In which case their highlights would still not be defined
> > > 
> > > 
> > > Obviously the best solution would be to have a “setup_highlights” method you can call before generating the status line which would create all possible highlights but that is out of the scope of this PR, I gave you everything you need to solve it properly if you wish.
> > 
> > 
> > What about the solution that creates highlights asynchronously? Does that work?
> 
> I’m not sure which solution you mean, I tried both `schedule_wrap` and async loop and they didn’t work.
> 
> I still believe this is the best and simplest solution, even if later down the life of the session there will be a dynamically generated highlight it doesn’t matter, all we’re trying to solve here is the intro screen, calling the generator twice seems to solve it as the other dynamically generated highlights will usually happen in a file buffer at which point you lost the intro screen anyways.

I think this is an issue that should be reported to neovim upstream since this is not really the fault of the plugins. So IMHO it's better if you make an issue upstream rather than creating a hacky workaround

ibhagwan:
A neovim issue might take years to get into main, I wouldn’t consider this too hacky, solving this issue for the price of 2 status line generation calls at init seems like a decent trade to me.

famiu:
> A neovim issue might take years to get into main, I wouldn’t consider this too hacky, solving this issue for the price of 2 status line generation calls at init seems like a decent trade to me.

But it's not even solving the issue for anything beyond the simple usecase. And it's not like Feline is the only plugin that suffers from those issues because of it. Even if it takes a while for the issue to be dealt with, I think it should be reported upstream

ibhagwan:
> > A neovim issue might take years to get into main, I wouldn’t consider this too hacky, solving this issue for the price of 2 status line generation calls at init seems like a decent trade to me.
> 
> But it's not even solving the issue for anything beyond the simple usecase. And it's not like Feline is the only plugin that suffers from those issues because of it. Even if it takes a while for the issue to be dealt with, I think it should be reported upstream

While I agree this should be solved upstream if possible (although from what I read the highlight forces a redraw so it might not be solve-able easily or might not be worth it at all), when it comes to feline the simple use case (intro screen) is all it aims to solve and it does, and I think for a fairly “low cost”, I see no harm in running the generator twice to solve this.

Anyways, it’s your code do with it as you wish, you tagged this bug “help wanted” which I did, if you’re not in agreement on the solution or have no intention to fix it in a way that you feel better about feel free to close this PR. I personally found this behavior annoying (don’t ask me why lol) and will continue to use my own fork of feline and rebase it moving forward.

famiu:
> > > A neovim issue might take years to get into main, I wouldn’t consider this too hacky, solving this issue for the price of 2 status line generation calls at init seems like a decent trade to me.
> > 
> > 
> > But it's not even solving the issue for anything beyond the simple usecase. And it's not like Feline is the only plugin that suffers from those issues because of it. Even if it takes a while for the issue to be dealt with, I think it should be reported upstream
> 
> While I agree this should be solved upstream if possible (although from what I read the highlight forces a redraw so it might not be solve-able easily or might not be worth it at all), when it comes to feline the simple use case (intro screen) is all it aims to solve and it does, and I think for a fairly “low cost”, I see no harm in running the generator twice to solve this.
> 
Neither do I. All I'm saying is that it's not a proper or universal fix. And that an issue should be made upstream even if this pull request is merged.
> Anyways, it’s your code do with it as you wish, you tagged this bug “help wanted” which I did, if you’re not in agreement on the solution or have no intention to fix it in a way that you feel better about feel free to close this PR. I personally found this behavior annoying (don’t ask me why lol) and will continue to use my own fork of feline and rebase it moving forward.
> 
The "help wanted" label was to help me determine what was causing the issue. But I suppose it's harmless to merge this PR. Either way, I really recommend making an upstream issue nonetheless.

I'll merge the PR now though. Sorry for any misunderstandings caused.

ibhagwan:
Ty!

famiu:
I had to revert this PR since it was apparently causing issues. Turns it the fix wasn't harmless after all. Sorry but I suppose you'd have to get the issue fixed upstream after all

PR #34, opened 2021-08-21
Add icon field to all components

### LINKED ISSUE #28: [Feature] Allow specific highlighting for icons
Is it possible to set a different highlighting for a component's icon?

If not I think it would be great if their highlighting were configurable just like how the separators are configured. I.e. something like this:

```lua
table.insert(components.right.active, {
  provider = 'git_branch',
  hl = { bg = 'line_bg' },
  icon = {
    str = '  '
    hi = {fg = 'orange'}
  }
})

```

ram02z:
Closes #28

I took a stab at trying add `icon` as field for any component. Due to the fact that the builtin providers have fallback icons, I needed to introduce an optional return value for all builtin providers. This was probably the easiest solution that wouldn't require refactoring how providers work in feline. 

Additionally, `icon` can accept a table. The fields accepted are exactly the same as the fields accepted for the separators.

I also refactored  some of the LSP providers. `lsp_client_names` no longer has an icon after every attached client so it can benefit from the new fields added to `icon`. `diagnostic_`* functions now check if LSP is attached or if no diagnostics are available.

I tested all the builtin providers and so far everything works as expected. I will add docs once the PR is reviewed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,7 +244,7 @@ enabled = function()
 end
 ```
 
-- `icon` (table or string): Some inbuilt providers like `git_branch`, provide default icons. If you either don't have a patched font or don't like the default icon that Feline provides, you may set this value to use any icon you want instead. Additionally, you can also change the highlight specifically for the icon. To do this you need to pass a table containing `str` and `hl`, where `str` would represent the icon and `hl` would represent the icon highlight. The icons's highlight works just like the `hl` component's values. For example:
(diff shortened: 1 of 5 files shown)


famiu:
@ram02z the PR seems fine but if possible you should also try to add documentation for it in README.md so I can finally merge the PR.

ram02z:
@famiu I have added documentation for it to the README. This should be ready to merge now.

PR #39, opened 2021-08-31
Remove extra space in `git_diff_changed` provider

6cdh:
`\uf9c9` (the default icon of `git_diff_changed` provider) in nerd font is larger (2 spaces) than it is expected (1 space). It can be confirmed in kitty, konsole, and hyper.

This makes it looks weird in terminal.
I don't know why, but I removed a trailing space to make it look good.

Before:

image

After:

image

### DIFF
--- a/lua/feline/providers/git.lua
+++ b/lua/feline/providers/git.lua
@@ -49,7 +49,7 @@ function M.git_diff_changed(component)
     local str = ''
 
     if gsd and gsd['changed'] and gsd['changed'] > 0 then
-        icon = component.icon or ' 柳 '
+        icon = component.icon or ' 柳'
         str = str .. gsd.changed
     end
 


famiu:
But it's also the case for `git_dif_added` and `git_diff_removed`

6cdh:
`git_diff_added` and `git_diff_removed` look good. This is a experiment in kitty:

image

`\uf9c9` is larger than others. I don't know why. Can you reproduce it?

6cdh:
Maybe because \uF900 - \uFAFF is range of CJK Compatibility Ideographs. CJK characters are wider than Latin alphabet.
So we don't need to change `git_diff_added` and `git_diff_removed`.

famiu:
Alright then, sounds reasonable enough

PR #40, opened 2021-09-01
Fix highlight background for `file_info`

PHSix:
I've found a new bug introduced in #38 . The file icon's background color automatically turns blacker with Normal when my feed background color is light.
before version:
front
fixed version:
fixed

### DIFF
--- a/lua/feline/providers/file.lua
+++ b/lua/feline/providers/file.lua
@@ -90,7 +90,8 @@ function M.file_info(component)
     local icon = component.icon
     if not icon then
         local ic, hl_group = require("nvim-web-devicons").get_icon(filename, extension, { default = true })
-        icon = { str = ic, hl = hl_group }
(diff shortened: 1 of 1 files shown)


numbcoder:
Thanks for fixing this bug.

But I think there is a better way to get highlight group colors. Too many `vim.fn.xxx` call will cause bad performance
```lua
local fg = string.format('#%06x', vim.api.nvim_get_hl_by_name(hl_group, true)['foreground'])
```

ram02z:
I agree with @numbcoder, his snippet works. However, the highlight background will not be fixed by omitting the bg. If you take a look at the `parse_hl` function, it inherits the `colors.bg` if `hl.bg` is not present, so the highlight background will still look like the following:
image

famiu:
> I have drafted a hacky solution, whereby `parse_hl` takes an extra parameter which is the parent_hl to be used by the `parse_icon` function.

A pull request would be welcome

PR #41, opened 2021-09-01
Refactor `parse_hl`

ram02z:
- `parse_hl` now accepts an optional parameter, which allows the highlight
to inherit parent highlight group.

- External highlight groups are now parsed using `get_hl`. If the highlight group doesn't exist, an empty string is returned (No highlighting will be performed).

- `get_hl` also parses styles, which means all attrs are checked manually. Not sure if there is a better method of mapping the cterm attributes.

Since this is refactors an important function, this could have adverse side effects. I have tested this on neovim-nightly using the default feline config.

@AllenDang Can you test this PR to see it resolves your issue?

See comment: https://github.com/famiu/feline.nvim/pull/40#issuecomment-910183007


famiu:
Why not just use the highlight name directly instead of using the get_hl function?

famiu:
This seems far too complicated imo. For most highlights get_hl isn't even necessary and you can just use the highlight name directly

ram02z:
> 
> 
> This seems far too complicated imo. For most highlights get_hl isn't even necessary and you can just use the highlight name directly

Without the get_hl function, external highlight groups are returned instantly, which means the highlight group cannot inherit any missing values from the parent or the default.

famiu:
Maybe make a separate parse_icon_hl function that works differently and keep the old parse_hl?

ram02z:
@famiu Done! I also kept the `get_hl` function as it useful to inherit missing values from highlight groups. I could also add it to the original `parse_hl`, but the only benefit would be that `colors.bg` or `colors.fg` would be inherited if background or foreground is missing from a highlight group.

famiu:
Honestly I think get_hl should be removed. If someone's using the name of a highlight, they probably know what they're doing

famiu:
I'm wondering 2 things at this point.

1. If a person is using a highlight name without a bg, they probably intend to have it without a bg.. The whole point of using highlight names is to make sure Feline doesn't do anything to alter it.

2. What's the performance of this compared to the current solution?

ram02z:
> 
>     1. If a person is using a highlight name without a bg, they probably intend to have it without a bg.. The whole point of using highlight names is to make sure Feline doesn't do anything to alter it.

That is a fair point. I can remove `get_hl` then. I just thought it may solve @AllenDang's issue since his `Normal` hl group was missing the background, so I think it may not be so bad to override the group for icons (or items that have a parent object).

> 
>     2. What's the performance of this compared to the current solution?

Well, `get_hl` will only get called if the user passes a string to `icon.hl` and most of the complexity is due to `get_hl` parsing all additional styles.

I am fine with simplifying the PR and making `parse_icon_hl` the only addition, however, I do think my solution is more complete.

famiu:
> I am fine with simplifying the PR and making `parse_icon_hl` the only addition, however, I do think my solution is more complete.

I'd honestly insist not doing anything if the value of hl is a string and removing hl, and only changing things when it isn't a string. I was also thinking about caching get_hl results, but changes to the highlight would mess with that

famiu:
Upon seeing the code again, I realized parse_icon_hl was basically a duplicate of parse_hl since get_hl was removed, so I removed it and allowed parse_hl to replicate its functionality without altering pre-existing code. That being said, you might want to change the title of the pull request now that the function is removed. And lastly, it'd be nice if you take one last look at the code to see if everything's alright before I merge it.

ram02z:
> That being said, you might want to change the title of the pull request now that the function is removed. And lastly, it'd be nice if you take one last look at the code to see if everything's alright before I merge it.

I updated the PR title and tested the new changes successfully against the default feline config.

PR #42, opened 2021-09-02
perf(providers): reduce `vim.fn` calls

ram02z:
https://github.com/famiu/feline.nvim/pull/40#issuecomment-909975582 inspired me to replace as many `vim.fn` calls with either `vim.api` functions or vim statusline items (see http://vimdoc.sourceforge.net/htmldoc/options.html#'statusline').


famiu:
Thank you for your contribution

PR #43, opened 2021-09-02
feat(file_info): add read-only indicator

### LINKED ISSUE #35: Read only tag on the feline would be great
some times we start editing a file then if we try to save it it shows read only can you add  a read only tag in feline that would be great

famiu:
Closes #35

### DIFF
--- a/README.md
+++ b/README.md
@@ -918,9 +918,11 @@ The Vi-mode provider also provides a helper function `get_mode_highlight_name()`
 
 #### File Info
 
-The `file_info` provider has two special component values:
+The `file_info` provider has three special component values:
 - `file_modified_icon` (string): The icon that is shown when a file is modified.<br>
 Default:`'●'`
+- `file_readonly_icon` (string): The icon that is shown when a file is read-only.<br>
+Default:`'🔒'`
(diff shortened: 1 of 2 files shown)


famiu:
Wondering if the indicator should be a part of file_info as it is now in the PR or if I should modularize it into a separate component

ram02z:
>
> Wondering if the indicator should be a part of file_info as it is now in the PR or if I should modularize it into a separate component

I think you should make it a separate component as it will allow users to set custom highlighting for the provider.

famiu:
> > Wondering if the indicator should be a part of file_info as it is now in the PR or if I should modularize it into a separate component
> 
> I think you should make it a separate component as it will allow users to set custom highlighting for the provider.

While that's a valid point, the modified indicator is also a part of file_info and that doesn't seem to cause any issues. Making a separate component, while allowing higher customizability, will increase component clutter. So I'm not sure if the tradeoff is worth it

famiu:
After giving it some thought I'm going to go with putting it in file_info, it's just way simpler imo and I believe that showing whether the file is read-only or not is within the scope of the file_info provider.

PR #44, opened 2021-09-02
feat(file_info): add `colored_icon` option

ram02z:
https://github.com/famiu/feline.nvim/issues/37#issuecomment-909849267 wasn't a fan of colored icons by default, so perhaps this should be behind a setting. I am wondering if it should be the default behaviour though, and instead allow the user to disable the colored icons if they want to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -918,7 +918,9 @@ The Vi-mode provider also provides a helper function `get_mode_highlight_name()`
 
 #### File Info
 
-The `file_info` provider has two special component values:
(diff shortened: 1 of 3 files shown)


famiu:
Made a few changes to it. I think it should be true by default. You should review the changes to see if everything's fine

ram02z:
> 
> 
> Made a few changes to it. I think it should be true by default. You should review the changes to see if everything's fine

LGTM

PR #51, opened 2021-09-07
New format for components

### LINKED ISSUE #49: New format for components
Recently, I've been thinking of changing the format in which components are defined, which would obviously be a breaking change, though I plan to deprecate the current format first before removing it outright in order to give people a chance to switch.

Before, the component table had the left, mid and right tables inside of it and each of those tables had an 'active' table and an 'inactive' table inside of them. The new format would instead have the 'active' table and 'inactive' table directly inside the component table, with a variable amount of sections for the active and inactive statusline. So you're not limited to the 3 sections (left, mid and right) but instead can have 4 or even more sections. Another benefit is that it allows grouping active and inactive statusline elements together. Here's what it'd look like in code:

```lua
components = {
    -- active statusline
    active = {
        -- left section
        {
            -- first component
            {
                -- insert component code here
            },
            -- second component
            {
                -- insert component code here
            }
        },
        -- mid section
        {
            -- insert code for section components here
            -- similarly to what was shown above in the code for the left section
        },
        -- right section
        {
            -- insert code for section components here
            -- similarly to what was shown above in the code for the left section
        }
    },
    -- inactive statusline
    inactive = {
        -- insert sections and section components here
        -- similarly to what was shown above in the code for the active statusline
    }
}
```

I feel like this change to the component table layout will make configuration simpler and more organized, while also adding flexibility. But I'm not 100% sure if I should implement it yet since this is going to be a breaking change. So I'd like other opinions on it if possible.

famiu:
Closes #49 

Currently added support for new format and made presets use the new format. ~~Need to add a deprecation warning for people using the old format and change docs accordingly before merging this.~~ It's ready for merge! New format works fine for me. Though I'd appreciate if others were willing to give it a spin, tell me if it works fine and report any problems encountered.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,76 +133,89 @@ If you don't mind getting your hands dirty, then I recommend making your own sta
 
 #### Components
 
-Inside the `components` table, there needs to be three more tables, `left`, `mid` and `right`, which will dictate if the component should be put in the left side, middle or the right side of the statusline. And in each of those tables, there needs to be two more tables, `active` and `inactive`, which will dictate whether the component is a part of the statusline when it's in the active window or the inactive window.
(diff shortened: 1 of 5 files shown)


famiu:
I'm just going to merge this pull request since I've tested it for a bit and it seems to work fine

PR #55, opened 2021-09-08
fix: make statusline update in all windows

### LINKED ISSUE #52: [Bug] statusline is updated only in current window
**Neovim version**
NVIM v0.5.0
Build type: Release
LuaJIT 2.0.5

**Feline version**
8c809b2416fa6773719f3c391eb7b124d67c866e (latest as of now)

**Describe the bug**
Statusline is updated only in current window, so inactive statusline isn't updated. This is a problem because in the inactive part i put the window number (return value of `vim.fn.winnr()`) to make easier to jump between windows with a keybinding (space+number). When deleting a window, for example, window numbers aren't updated, making navigation between windows difficult.

**To Reproduce**
1. Vertical split (`<C-w>v`)
2. Vertical split (`<C-w>v`)
3. Go to first window (`<C-w>w`)
4. Delete it (`<C-w>c`)
5. The window on the right shows number 3, even if it is window 2.

If you jump to window 2 and come back to window 1 you'll see the number updated.

**Expected behavior**
Inactive windows should be updated.

**Screenshots**
After step 3:
st1

After step 4:
st2

famiu:
Closes #52

I'd like people to review it and check if everything works as intended before I merge this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,7 +221,7 @@ components.active[3][2] = {
 
 Now you can customize each component to your liking. Most values that a component requires can also use a function without arguments, with the exception of the `provider` value, which can take arguments (more about that below). Feline will automatically evaluate the function if it is given a function. But in case a function is provided, the type of value the function returns must be the same as the type of value required by the component. For example, since `enabled` requires a boolean value, if you set it to a function, the function must also return a boolean value. Note that you can omit all of the component values except `provider`, in which case the defaults would be used instead. A component can have the following values:
 
-- `provider` (string or function): If it's a string, it represents the text to show. If it's a function, it must return a string when called. As a function it may also optionally return an `icon` component alongside the string when called, which would represent the provider's icon, possibly along with the icon highlight group configuration. The function can take either no arguments, or one argument which would contain the component itself, or it can take two arguments, the component and the window number of the window for which the statusline is being generated.
+- `provider` (string or function): If it's a string, it represents the text to show. If it's a function, it must return a string when called. As a function it may also optionally return an `icon` component alongside the string when called, which would represent the provider's icon, possibly along with the icon highlight group configuration. The function can take either no arguments, or one argument which would contain the component itself, or it can take two arguments, the component and the window handler of the window for which the statusline is being generated.
 
 ```lua
 -- Provider that shows current line in file
@@ -238,14 +238,14 @@ provider = function(component)
     end
 end
 
--- Providers can also take the window number as an argument
-provider = function(component, winnr)
-    return (component.icon or '') .. tostring(vim.api.nvim_win_get_buf(winnr))
+-- Providers can also take the window handler as an argument
+provider = function(component, winid)
+    return (component.icon or '') .. tostring(vim.api.nvim_win_get_buf(winid))
 end
 
--- If you only need the window number, you can not avoid using the component value
-provider = function(_, winnr)
-    return vim.api.nvim_win_get_cursor(winnr)[1]
+-- If you only need the window handler, you can avoid using the component value like this:
+provider = function(_, winid)
+    return vim.api.nvim_win_get_cursor(winid)[1]
 end
 
 -- Providers can also simply just contain a string, such as:
--- a/README.md
+++ b/README.md
@@ -219,9 +219,9 @@ components.active[3][2] = {
 
 **NOTE:** If you use the index instead of table.insert, remember to put the correct index. Also keep in mind that unlike most other programming languages, Lua indices start at `1` instead of `0`.
 
-Now you can customize each component to your liking. Most values that a component requires can also use a function without arguments, with the exception of the `provider` value, which can take one argument, more about that below. Feline will automatically evaluate the function if it is given a function. But in case a function is provided, the type of value the function returns must be the same as the type of value required by the component. For example, since `enabled` requires a boolean value, if you set it to a function, the function must also return a boolean value. Note that you can omit all of the component values except `provider`, in which case the defaults would be used instead. A component can have the following values:
+Now you can customize each component to your liking. Most values that a component requires can also use a function without arguments, with the exception of the `provider` value, which can take arguments (more about that below). Feline will automatically evaluate the function if it is given a function. But in case a function is provided, the type of value the function returns must be the same as the type of value required by the component. For example, since `enabled` requires a boolean value, if you set it to a function, the function must also return a boolean value. Note that you can omit all of the component values except `provider`, in which case the defaults would be used instead. A component can have the following values:
 
-- `provider` (string or function): If it's a string, it represents the text to show. If it's a function, it must return a string when called. As a function it may also optionally return an `icon` component alongside the string when called, which would represent the provider's icon, possibly along with the icon highlight group configuration. The function can take either no arguments, or one argument which would contain the component itself.
+- `provider` (string or function): If it's a string, it represents the text to show. If it's a function, it must return a string when called. As a function it may also optionally return an `icon` component alongside the string when called, which would represent the provider's icon, possibly along with the icon highlight group configuration. The function can take either no arguments, or one argument which would contain the component itself, or it can take two arguments, the component and the window number of the window for which the statusline is being generated.
 
 ```lua
 -- Provider that shows current line in file
@@ -238,7 +238,17 @@ provider = function(component)
     end
 end
 
--- Providers can also just contain a simple string, such as:
+-- Providers can also take the window number as an argument
+provider = function(component, winnr)
+    return (component.icon or '') .. tostring(vim.api.nvim_win_get_buf(winnr))
+end
+
+-- If you only need the window number, you can not avoid using the component value
+provider = function(_, winnr)
+    return vim.api.nvim_win_get_cursor(winnr)[1]
+end
+
+-- Providers can also simply just contain a string, such as:
 provider = 'some text here'
 ```
 
--- a/README.md
+++ b/README.md
@@ -552,6 +552,8 @@ Now that we've learned to set up both the components table and the properties ta
 - `preset` - Set it to use a preconfigured statusline. Currently it can be equal to either `default` for the default statusline or `noicon` for the default statusline without icons. You don't have to put any of the other values if you use a preset, but if you do, your settings will override the preset's settings. To see more info such as how to modify a preset to build a statusline, see: [Modifying an existing preset](#3.-modifying-an-existing-preset)
 - `colors` - A table containing custom [color value presets](#value-presets).
 - `separators` - A table containing custom [separator value presets](#value-presets).
+- `update_triggers` - A list of autocmds that trigger an update of the statusline in inactive windows.<br>
+Default: `{'VimEnter', 'WinEnter', 'WinClosed', 'FileChangedShellPost'}`
 - `components` - The components table.
 - `properties` - The properties table.
 - `vi_mode_colors` - A table containing colors associated with Vi modes. It can later be used to get the color associated with the current Vim mode using `require('feline.providers.vi_mode').get_mode_color()`. For more info on it see the [Vi-mode](#vi-mode) section.<br><br>Here is a list of all possible vi_mode names used with the default color associated with them:
--- a/lua/feline/defaults.lua
+++ b/lua/feline/defaults.lua
@@ -58,4 +58,11 @@ M.vi_mode_colors = {
     ['NONE'] = 'yellow'
 }
 
+M.update_triggers = {
+    'VimEnter',
+    'WinEnter',
+    'WinClosed',
+    'FileChangedShellPost'
+}
+
 return M
--- a/lua/feline/generator.lua
+++ b/lua/feline/generator.lua
@@ -140,12 +140,12 @@ local function parse_icon(icon, parent_hl)
 end
 
 -- Parse component provider
-local function parse_provider(provider, component)
+local function parse_provider(provider, component, winnr)
     local icon
     if type(provider) == "string" and type(providers[provider]) == "function" then
-        provider, icon = providers[provider](component)
+        provider, icon = providers[provider](component, winnr)
     elseif type(provider) == "function" then
-        provider, icon = provider(component)
+        provider, icon = provider(component, winnr)
     end
 
     if type(provider) ~= "string" then
@@ -158,18 +158,17 @@ local function parse_provider(provider, component)
 end
 
 -- Parses a component alongside its highlight
-local function parse_component(component)
+local function parse_component(component, winnr)
     local enabled = evaluate_if_function(component.enabled, true)
 
     if not enabled then return '' end
-
 
     local hl = evaluate_if_function(component.hl, {})
 
     local left_sep_str = parse_sep_list(component.left_sep, hl.bg)
     local right_sep_str = parse_sep_list(component.right_sep, hl.bg)
 
-    local str, icon = parse_provider(component.provider, component)
+    local str, icon = parse_provider(component.provider, component, winnr)
 
     local hlname = parse_hl(hl)
 
@@ -184,12 +183,12 @@ end
 
 -- Parse components of a section of the statusline
 -- (For old component table format)
-local function parse_statusline_section_old(section, type)
+local function parse_statusline_section_old(section, type, winnr)
     if feline.components[section] and feline.components[section][type] then
         local section_components = {}
 
         for _, v in ipairs(feline.components[section][type]) do
-            section_components[#section_components+1] = parse_component(v)
+            section_components[#section_components+1] = parse_component(v, winnr)
         end
 
         return table.concat(section_components)
@@ -199,25 +198,25 @@ local function parse_statusline_section_old(section, type)
 end
 
 -- Parse components of a section of the statusline
-local function parse_statusline_section(section)
+local function parse_statusline_section(section, winnr)
     local components = {}
 
     for _, component in ipairs(section) do
-        components[#components+1] = parse_component(component)
+        components[#components+1] = parse_component(component, winnr)
     end
 
     return table.concat(components)
 end
 
 -- Generate statusline by parsing all components and return a string
-function M.generate_statusline(is_active)
+function M.generate_statusline(winnr)
     if not feline.components then
         return ''
     end
 
     local statusline_type
 
-    if is_active and not is_forced_inactive() then
+    if winnr == vim.api.nvim_get_current_win() and not is_forced_inactive() then
         statusline_type='active'
     else
         statusline_type='inactive'
@@ -229,16 +228,16 @@ function M.generate_statusline(is_active)
         local sections = {}
 
         for _, section in ipairs(feline.components[statusline_type]) do
-            sections[#sections+1] = parse_statusline_section(section)
+            sections[#sections+1] = parse_statusline_section(section, winnr)
         end
 
         return table.concat(sections, '%=')
     else
         return string.format(
             "%s%%=%s%%=%s",
-            parse_statusline_section_old('left', statusline_type),
-            parse_statusline_section_old('mid', statusline_type),
-            parse_statusline_section_old('right', statusline_type)
+            parse_statusline_section_old('left', statusline_type, winnr),
+            parse_statusline_section_old('mid', statusline_type, winnr),
+            parse_statusline_section_old('right', statusline_type, winnr)
         )
     end
 end
--- a/lua/feline/generator.lua
+++ b/lua/feline/generator.lua
@@ -140,12 +140,12 @@ local function parse_icon(icon, parent_hl)
 end
 
 -- Parse component provider
-local function parse_provider(provider, component, winnr)
+local function parse_provider(provider, component, winid)
     local icon
     if type(provider) == "string" and type(providers[provider]) == "function" then
-        provider, icon = providers[provider](component, winnr)
+        provider, icon = providers[provider](component, winid)
     elseif type(provider) == "function" then
-        provider, icon = provider(component, winnr)
+        provider, icon = provider(component, winid)
     end
 
     if type(provider) ~= "string" then
@@ -158,7 +158,7 @@ local function parse_provider(provider, component, winnr)
 end
 
 -- Parses a component alongside its highlight
-local function parse_component(component, winnr)
+local function parse_component(component, winid)
     local enabled = evaluate_if_function(component.enabled, true)
 
     if not enabled then return '' end
@@ -168,7 +168,7 @@ local function parse_component(component, winnr)
     local left_sep_str = parse_sep_list(component.left_sep, hl.bg)
     local right_sep_str = parse_sep_list(component.right_sep, hl.bg)
 
-    local str, icon = parse_provider(component.provider, component, winnr)
+    local str, icon = parse_provider(component.provider, component, winid)
 
     local hlname = parse_hl(hl)
 
@@ -183,12 +183,12 @@ end
 
 -- Parse components of a section of the statusline
 -- (For old component table format)
-local function parse_statusline_section_old(section, type, winnr)
+local function parse_statusline_section_old(section, type, winid)
     if feline.components[section] and feline.components[section][type] then
         local section_components = {}
 
         for _, v in ipairs(feline.components[section][type]) do
-            section_components[#section_components+1] = parse_component(v, winnr)
+            section_components[#section_components+1] = parse_component(v, winid)
         end
 
         return table.concat(section_components)
@@ -198,25 +198,25 @@ local function parse_statusline_section_old(section, type, winnr)
 end
 
 -- Parse components of a section of the statusline
-local function parse_statusline_section(section, winnr)
+local function parse_statusline_section(section, winid)
     local components = {}
 
     for _, component in ipairs(section) do
-        components[#components+1] = parse_component(component, winnr)
+        components[#components+1] = parse_component(component, winid)
     end
 
     return table.concat(components)
 end
 
 -- Generate statusline by parsing all components and return a string
-function M.generate_statusline(winnr)
+function M.generate_statusline(winid)
     if not feline.components then
         return ''
     end
 
     local statusline_type
 
-    if winnr == vim.api.nvim_get_current_win() and not is_forced_inactive() then
+    if winid == vim.api.nvim_get_current_win() and not is_forced_inactive() then
         statusline_type='active'
     else
         statusline_type='inactive'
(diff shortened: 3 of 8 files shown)


ram02z:
If you start telescope (in insert mode) and open a file, the windows statusline doesn't get updated and the vi provider is stuck on insert mode. I can also reproduce with nnn.vim, following the same steps, but instead the vi provider is stuck on terminal mode. Also, floating windows trigger an inactive and empty statusline, not sure if thats intended. 

Strange behaviour when you have a split open with files of different filetypes, I think this shows that the wrong window is getting updated.

Reproduction steps:
1. Open two files of two different filetypes
2. Create a split to force one window to have an inactive statusline
3. The unfocused window will show the filetype of the focused file (this is using the default config).

xeluxee:
It looks like feline is updated in every window but with active window data.

st1
As you can see in the image, if I go to NvimTree every window shows NvimTree and window number 1

In the same way if I go on split 2 every window gets its infos from the second window
st2

famiu:
@xeluxee @ram02z So I'm trying to edit the providers so they can take the window number and use the appropriate buffer to show their info. But an issue I came across while doing that is related to the Git provider. Git uses the gitsigns_status_dict of gitsigns.nvim, which it only seems to provide for the current buffer. So it wouldn't be possible to use it for inactive windows, unless there's a way. Thoughts?

ram02z:
I was playing around the code, and I think the only way to ensure that the win number is correct is to make it a named argument for function providers.

famiu:
> I was playing around the code, and I think the only way to ensure that the win number is correct is to make it a named argument for function providers.

That is what I'm doing. And I'm actually just done with that. Will make a separate PR for it in a bit, just need to do some final cleanups

xeluxee:
Give a look at [`nvim_win_call`](https://neovim.io/doc/user/api.html#nvim_win_call())
It lets you execute a function as it was in a specific window.

famiu:
@ram02z @xeluxee Okay so I'm done with stuff. But there's one issue, I need to use the gitsigns_status_dict and gitsigns_head buffer values of gitsigns.nvim for the git providers. In order to do that, I used `vim.api.nvim_buf_get_var`. But the problem with that is, gitsigns doesn't immediately define the dicts after startup, and if the dict doesn't exist, `vim.api.nvim_buf_get_var` shows an error, whereas previously I used `vim.b.gitsigns_status_dict` directly which just returned nil if it wasn't defined, which was easy to check. Any ideas on how to work around that?

ram02z:
You could wrap it in a pcall, kind of like this:
```lua
local ok, value = pcall(vim.api.nvim_buf_get_var, 0, "...")
if not ok then
  return
end
```

famiu:
> You could wrap it in a pcall, kind of like this:
> 
> ```lua
> local ok, value = pcall(vim.api.nvim_buf_get_var, 0, "...")
> if not ok then
>   return
> end
> ```

Tried that already, doesn't work

ram02z:
`vim.fn.getbufvar` returns an empty string when the variable is not found.

famiu:
> You could wrap it in a pcall, kind of like this:
> 
> ```lua
> local ok, value = pcall(vim.api.nvim_buf_get_var, 0, "...")
> if not ok then
>   return
> end
> ```

That's what I tried at first, but it didn't work. Then after seeing your code I realized I made a silly mistake which was causing it to not work, but it works now!

famiu:
Merging #56 would be needed before I can make further progress on this, so it'd be nice if that was tested out.

xeluxee:
@famiu just wondering why not using [`nvim_win_call`](https://neovim.io/doc/user/api.html#nvim_win_call())
It would be simpler, there wouldn't be any need to edit current providers...

famiu:
> @famiu just wondering why not using [`nvim_win_call`](https://neovim.io/doc/user/api.html#nvim_win_call())
> It would be simpler, there wouldn't be any need to edit current providers...

well, let me see if I can do that, then

famiu:
> @famiu just wondering why not using [`nvim_win_call`](https://neovim.io/doc/user/api.html#nvim_win_call())
> It would be simpler, there wouldn't be any need to edit current providers...

How would we use it to call providers that take an argument? As far as I can see, it only works for functions that don't take any arguments

famiu:
I incorporated the changes in #56 in this PR and closed #56 since it's not usable by itself and needs the changes in this PR to function properly.

famiu:
First problem I noticed: the cursor doesn't update and I don't have any idea why. Need to look into it more

famiu:
> First problem I noticed: the scroll indicator doesn't update and I don't have any idea why. Need to look into it more

It seems that it's not unique to the scroll indicator, none of the providers except the file position are updated, probably because the file position provider uses Neovim's built-in statusline component variables. This indicates that the generator is only being called once when the buffer is opened.

famiu:
I managed to fix the components not being updated

famiu:
It seems some components (such as the git branch name) are still not being updated, need to look into it more

xeluxee:
> > @famiu just wondering why not using [`nvim_win_call`](https://neovim.io/doc/user/api.html#nvim_win_call())
> > It would be simpler, there wouldn't be any need to edit current providers...
> 
> How would we use it to call providers that take an argument? As far as I can see, it only works for functions that don't take any arguments

@famiu this is a sample code using [`nvim_win_call`](https://neovim.io/doc/user/api.html#nvim_win_call())
``` lua
local myfunc = function (add, sub)
  print(string.format("Window number %d has id %d: value %d", vim.fn.winnr(), vim.fn.win_getid(), vim.fn.winnr()+add-sub))
end

for winnr, winid in pairs(vim.api.nvim_list_wins()) do
  vim.api.nvim_win_call(winid, function() myfunc(5, -5) end)
end
```

Output:
```
Window number 1 has id 1008: value 11                                                                                                                                         
Window number 2 has id 1000: value 12 
```

xeluxee:
> I managed to fix the components not being updated. This seems to be ready for merge now. Just need people to review it

I'm using 4727b3535a43832d088e70f525a17704beb159eb and this problem remains: every window takes data from current window
st1

famiu:
> > I managed to fix the components not being updated. This seems to be ready for merge now. Just need people to review it
> 
> I'm using [4727b35](https://github.com/famiu/feline.nvim/commit/4727b3535a43832d088e70f525a17704beb159eb) and this problem remains: every window takes data from current window
> st1

Cannot reproduce the issue. Does this work for you @ram02z?

ram02z:
> 
> Cannot reproduce the issue. Does this work for you @ram02z?

I can reproduce if I use a provider that returns `vim.fn.winnr()`. However, it updates fine when I use the `winnr` parameter. However, the `winnr` parameter actually returns the window handle, not number, which I guess won't solve @xeluxee 's issue. 
`nvim_list_wins` returns a table of window handles.

famiu:
> I can reproduce if I use a provider that returns `vim.fn.winnr()`. However, it updates fine when I use the `winnr` parameter.
> However, the `winnr` parameter actually returns the window handle, not number, which I guess won't solve @xeluxee 's issue.

Just use `nvim_win_get_number`

ram02z:
Ok since we have decided to take the reliable approach, I would also like to see a few of the changes made in #58 be introduced to this PR.

1. Update only inactive windows. There is no reason to update the active window because we reset the local statusline of the current window to use the global statusline for it anyway.
2. Filter out windows that we don't care about like floating windows and the current window.
3. Iterate over the tabpage windows not all the windows, since not all the windows are visible.
3. This is probably the most important change and it is to reduce the amount of autocmds we use to trigger the update of inactive windows. The reason I say this is that firstly we should avoid unnecessary redraws when possible. These unnecessary redraws can cause unforeseen issues like the one with we noticed with lsp hover. The culprit autocmd was `BufEnter` and by introducing it, we are also introducing a regression that was fixed upstream https://github.com/neovim/neovim/pull/10112. I have already found the bare minimum autocmds; VimEnter, WinEnter, FileChangedShellPost.

Another change unrelated to #58 is quite simple, change references of `winnr` to `winid` because we aren't actually providing the window number. 

Here is the patch I am using currently that builds on this PR based on my above comments: https://gist.github.com/ram02z/59761a0acb7fb86bccf71f84bc0c955d

famiu:
> Ok since we have decided to take the reliable approach, I would also like to see a few of the changes made in #58 be introduced to this PR.
> 
>     1. Update only inactive windows. There is no reason to update the active window because we reset the local statusline of the current window to use the global statusline for it anyway.
> 
>     2. Filter out windows that we don't care about like floating windows and the current window.
> 
>     3. Iterate over the tabpage windows not all the windows, since not all the windows are visible.
> 
>     4. This is probably the most important change and it is to reduce the amount of autocmds we use to trigger the update of inactive windows. The reason I say this is that firstly we should avoid unnecessary redraws when possible. These unnecessary redraws can cause unforeseen issues like the one with we noticed with lsp hover. The culprit autocmd was `BufEnter` and by introducing it, we are also introducing a regression that was fixed upstream [api/buffer: avoid spurios Buf[Win]Enter in nvim_buf_set_lines neovim/neovim#10112](https://github.com/neovim/neovim/pull/10112). I have already found the bare minimum autocmds; VimEnter, WinEnter, FileChangedShellPost. Working our way up is more reliable in my opinion.
> 
> 
> Another change unrelated to #58 is quite simple, change references of `winnr` to `winid` because we aren't actually providing the window number.
> 

I had actually already made all of the changes you listed before, just didn't push them yet. I've pushed them now

famiu:
Also I've tested this for a while, it seems to work fine for me. I didn't notice any slowdowns either so for all intents and purposes the performance seems to be about the same, which should be the case since inactive windows are updated only when the autocmd triggers. I'm wondering whether to merge this or wait for further testing

ram02z:
I think the only side effects this PR may have is that the someone may find a scenario whereby the inactive statusline isn't updating, but I haven't noticed any issues either, so if a scenario does arise, it will most likely be solved by adding the appropriate autocmd.

famiu:
> I think the only side effects this PR may have is that the someone may find a scenario whereby the inactive statusline isn't updating, but I haven't noticed any issues either, so if a scenario does arise, it will most likely be solved by adding the appropriate autocmd.

With that considered, I'm going to merge this then

xeluxee:
I can confirm this PR works well:
- providers in inactive windows works well thanks to `winid` parameter
- LSP hover doesn't make feline inactive in current window

But NvimTree problem remains, and it's independent from `splitright` because in another scenario inactive windows aren't updated.
To reproduce (no matter of `splitright` value):
1. Open a buffer fullscreen
2. Split vertically
3. Launch `:NvimTreeToggle`
4. Go to middle window
5. Launch `:NvimTreeToggle`
6. The number in the other window doesn't get updated
st1
_As you can see, there are only two windows but the right one shows number 3_

I tried to use some autocommand like `WinNew` or `WinClosed`, but they don't work...

famiu:
> I can confirm this PR works well:
> 
>     * providers in inactive windows works well thanks to `winid` parameter
> 
>     * LSP hover doesn't make feline inactive in current window
> 
> 
> But NvimTree problem remains, and it's independent from `splitright` because in another scenario inactive windows aren't updated.
> To reproduce (no matter of `splitright` value):
> 
>     1. Open a buffer fullscreen
> 
>     2. Split vertically
> 
>     3. Launch `:NvimTreeToggle`
> 
>     4. Go to middle window
> 
>     5. Launch `:NvimTreeToggle`
> 
>     6. The number in the other window doesn't get updated
>        st1
>        _As you can see, there are only two windows but the right one shows number 3_
> 
> 
> I tried to use some autocommand like `WinNew` or `WinClosed`, but they don't work...

Can you try adding the `BufWinEnter` autocmd and see if it works

ram02z:
>
> But NvimTree problem remains, and it's independent from `splitright` because in another scenario inactive windows aren't updated.
> To reproduce (no matter of `splitright` value):
> 
>     1. Open a buffer fullscreen
> 
>     2. Split vertically
> 
>     3. Launch `:NvimTreeToggle`
> 
>     4. Go to middle window
> 
>     5. Launch `:NvimTreeToggle`
> 
>     6. The number in the other window doesn't get updated
>        st1
>        _As you can see, there are only two windows but the right one shows number 3_
> 
> 
> I tried to use some autocommand like `WinNew` or `WinClosed`, but they don't work...

I can also reproduce this issue. However, I cannot reproduce if I skip the 4th step and also I cannot repoduce when using a plugin that also creates a left split, undotree, `UndoTreeToggle`.

famiu:
> > But NvimTree problem remains, and it's independent from `splitright` because in another scenario inactive windows aren't updated.
> > To reproduce (no matter of `splitright` value):
> > ```
> > 1. Open a buffer fullscreen
> > 
> > 2. Split vertically
> > 
> > 3. Launch `:NvimTreeToggle`
> > 
> > 4. Go to middle window
> > 
> > 5. Launch `:NvimTreeToggle`
> > 
> > 6. The number in the other window doesn't get updated
> >    st1
> >    _As you can see, there are only two windows but the right one shows number 3_
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I tried to use some autocommand like `WinNew` or `WinClosed`, but they don't work...
> 
> I can also reproduce this issue. However, I cannot reproduce if I skip the 4th step and also I cannot repoduce when using a plugin that also creates a left split, undotree, `UndoTreeToggle`.

Can you try adding `BufWinEnter` autocmd and see if it works

ram02z:
>
> Can you try adding `BufWinEnter` autocmd and see if it works

BufWinEnter doesn't work and also breaks lsp hover. WinLeave actually gets called at step 5 but the number still doesnt update.

famiu:
> > Can you try adding `BufWinEnter` autocmd and see if it works
> 
> BufWinEnter doesn't work and also breaks lsp hover. WinLeave actually gets called at step 5 but the number still doesnt update.

At this point I'm thinking if it's an issue with the window number itself. Can you try switching to a different window and switching back to see if it gets fixed or if it still shows the wrong numbers

ram02z:
> 
> At this point I'm thinking if it's an issue with the window number itself. Can you try switching to a different window and switching back to see if it gets fixed or if it still shows the wrong numbers

It is an issue with the window number. I took a look at how nvim-tree's handles window toggles and it uses `nvim_win_hide`, which makes sense why both `BufHidden` and `WinClosed` work. However, what I don't understand is why the window number isn't updated after the autocmd is triggered (`update_inactive_windows` definitely still gets called)

https://github.com/kyazdani42/nvim-tree.lua/blob/71122d798482e30c599d78aa7ae4a756c6e81a79/lua/nvim-tree/view.lua#L337

famiu:
> > At this point I'm thinking if it's an issue with the window number itself. Can you try switching to a different window and switching back to see if it gets fixed or if it still shows the wrong numbers
> 
> It is an issue with the window number. I took a look at how nvim-tree's handles window toggles and it uses `nvim_win_hide`, which makes sense why both `BufHidden` and `WinClosed` work. However, what I don't understand is why the window number isn't updated after the autocmd is triggered (`update_inactive_windows` definitely still gets called)
> 
> https://github.com/kyazdani42/nvim-tree.lua/blob/71122d798482e30c599d78aa7ae4a756c6e81a79/lua/nvim-tree/view.lua#L337

Maybe it's because the window numbers don't actually change if the buffer is hidden.

ram02z:
> 
> Maybe it's because the window numbers don't actually change if the buffer is hidden.

I mean I checked the window handle's number before and after step 5 and it does infact change from 3->2.

famiu:
> > Maybe it's because the window numbers don't actually change if the buffer is hidden.
> 
> I mean I checked the window handle's number before and after step 5 and it does infact change from 3->2.

Then it's very strange for the window number to not be updated even after `update_inactive_windows` is called

ram02z:
>
> Then it's very strange for the window number to not be updated even after `update_inactive_windows` is called. Either way I should probably add `BufHidden` as a trigger. Thoughts on that?

WinClosed seems more appropriate. It is very strange because if you add an autocmd like `CursorMoved` or `CursorHold` which update quite frequently, the window number does change to 2.

famiu:
> > Then it's very strange for the window number to not be updated even after `update_inactive_windows` is called. Either way I should probably add `BufHidden` as a trigger. Thoughts on that?
> 
> WinClosed seems more appropriate. It is very strange because if you add an autocmd like `CursorMoved` or `CursorHold` which update quite frequently, the window number does change to 2.

I realized that no matter how many autocmds we add, some functionality is still inevitably going to break due to information not being updated. So at one point we have to decide what to add and what not to. That or we can go fully inefficient by integrating `update_all_windows` inside `generate_statusline`, which'd mean the statusline would get updated all the time. Again, it's a tradeoff.

ram02z:
> 
> I realized that no matter how many autocmds we add, some functionality is still inevitably going to break due to information not being updated. So at one point we have to decide what to add and what not to. That or we can go fully inefficient by integrating `update_all_windows` inside `generate_statusline`, which'd mean the statusline would get updated all the time. Again, it's a tradeoff.

Yeah that is true. I think that the tradeoff is simply not worth it because I don't think most people want constant updates on their inactive windows(I know I don't). The current approach works fine but breaks with `nvim-tree` for some reason, I think that the issue should be tracked until an appropriate workaround/fix is found or you could leave this PR open. I don't think this will break any existing functionality though so maybe merging to master is best since we can get people to file bug reports if other plugins also have the same issue as `nvim-tree`.

famiu:
> > I realized that no matter how many autocmds we add, some functionality is still inevitably going to break due to information not being updated. So at one point we have to decide what to add and what not to. That or we can go fully inefficient by integrating `update_all_windows` inside `generate_statusline`, which'd mean the statusline would get updated all the time. Again, it's a tradeoff.
> 
> Yeah that is true. I think that the tradeoff is simply not worth it because I don't think most people want constant updates on their inactive windows(I know I don't). The current approach works fine but breaks with `nvim-tree` for some reason, I think that the issue should be tracked until an appropriate workaround/fix is found or you could leave this PR open. I don't think this will break any existing functionality though so maybe merging to master is best since we can get people to file bug reports if other plugins also have the same issue as `nvim-tree`.
> 
> I do think `WinClosed` should be added since it is emitted whenever `nvim_win_hide` or `nvim_win_close` is called.

We could have a configuration option that makes the statusline update constantly, though I don't know if that's worth adding. But yes, I will add WinClosed. And I'm going to merge this PR once we decide if we should add an option to update the statusline continuously or not and take appropriate actions for that.

ram02z:
> 
> We could have a configuration option that makes the statusline update constantly, though I don't know if that's worth adding. But yes, I will add `WinClosed`. And I'm going to merge this PR once we decide if we should add an option to update the statusline continuously or not and take appropriate actions for that. On another note, maybe allow users to configure the autocmds that trigger an inactive statusline update?

You know what, it might be better to leave it the end user to decide what additional autocmds they want to use. However, the current autocmds should stay as the default, and then only append autocmds that aren't default if a user specifies them. This approach does resonate with the design approach of feline.

xeluxee:
Just discovered sth interesting, this issue could be linked to NvimTree itself

```
:autocmd WinClosed                                                                                                                                                            
--- Autocommands ---                                                                                                                                                          
WinClosed                                                                                                                                                                     
    *         lua require('core.nvimtree').on_close()                                                                                                                         
feline  WinClosed                                                                                                                                                             
    *         lua require("feline").update_inactive_windows()                                                                                                                 
NvimTree  WinClosed                                                                                                                                                           
    *         lua require'nvim-tree'.on_leave()                                                                                                                               
bufferline_update  WinClosed                                                                                                                                                  
    *         call bufferline#update_async()
```

Maybe `on_leave()` is causing a wrong behaviour...

xeluxee:
Just implemented an update with debug method:

``` lua
function M.update_inactive_windows(debug)
    local current_win = api.nvim_get_current_win()

    for _, winid in ipairs(api.nvim_tabpage_list_wins(0)) do
        if debug == true then
            print(string.format("CHECKING window %d, while cur is %d", winid, current_win))
        end
        if api.nvim_win_get_config(winid).relative == '' and winid ~= current_win
        then
            if debug == true then
                print("-> Updated")
            end
            vim.wo[winid].statusline = require('feline').statusline(winid)
        end
    end

    -- Reset local statusline of current window to use the global statusline for it
    vim.wo.statusline = nil
end
```

I noticed that this function is called before NvimTree window is closed. Steps to reproduce:
1. Add this to autocommands in `init.lua`:
``` lua
        {
            'WinClosed',
            '*',
            'lua require("feline").update_inactive_windows(true)'
        },
```
2. `:set cmdheight=1` (so vim will stop to show messages)
3. restart and open NvimTree, then close it.
4. You'll see that messages are displayed before window is closed

This is why other windows don't get updated, because the first window (NvimTree) is still opened when `update_inactive_windows` is called.

ram02z:
Nice find @xeluxee! I remember seeing someone use `timer_start()` to defer execution until all other autocommands are run. I think `vim.loop.new_timer()` is the appropriate equivalent. I have tested the below code and I cannot reproduce your issue anymore. Let me know if it breaks anything :D

```lua
-- Uses a timer with value 0 to defer executing the function until after
-- all other autocommands have run. This will ensure that inactive windows
-- are updated after any changes.
function M.update_inactive_windows()
    local timer = vim.loop.new_timer()
    timer:start(0, 0, vim.schedule_wrap(function()
        local current_win = api.nvim_get_current_win()

        for _, winid in ipairs(api.nvim_tabpage_list_wins(0)) do
            if api.nvim_win_get_config(winid).relative == '' and winid ~= current_win
            then
                vim.wo[winid].statusline = require('feline').statusline(winid)
            end
        end

        -- Reset local statusline of current window to use the global statusline for it
        vim.wo.statusline = nil
    end))
end
```

famiu:
> UPDATE: seems like it breaks `vi_mode` on inactive windows. A quick reproduce is to invoke `:Telescope` and you should see Insert mode on all inactive windows (you need to have vi_mode on inactive windows). I don't think `vi_mode` should work on inactive windows anyway, so I guess it is fine?

That's the intended behavior though, right? `vi_mode` is the same for all windows since it's a global value independent of the window.

ram02z:
> 
> That's the intended behavior though, right? `vi_mode` is the same for all windows since it's a global value independent of the window.

Yeah that is true, but I feel like `vim.defer_fn` may have other unwarranted side effects even though it solves the nvim-tree issue. I am honestly not sure how else we can ensure that `update_inactive_windows` is executed after other autocmds.

famiu:
> > That's the intended behavior though, right? `vi_mode` is the same for all windows since it's a global value independent of the window.
> 
> Yeah that is true, but I feel like `vim.defer_fn` may have other unwarranted side effects even though it solves the nvim-tree issue. I am honestly not sure how else we can ensure that `update_inactive_windows` is executed after other autocmds.

I suppose this is still the best way we have. Another possibility I see is continuously updating inactive windows but in the background asynchronously. That being said, if we go with the current method, do you still think autocmd triggers for Feline updates should be user configurable?

ram02z:
>
> I suppose this is still the best way we have. Another possibility I see is continuously updating inactive windows but in the background asynchronously. That being said, if we go with the current method, do you still think autocmd triggers for Feline updates should be user configurable?

I think asynchronous behaviour in neovim is hard to get right but if you do have a solution, I don't mind reviewing it. 

I don't think there is any harm in making autocmds user configurable, since if it breaks, it will be on the end user and their configuration. Having the option to extend the autocmds sounds nice.

famiu:
> > I suppose this is still the best way we have. Another possibility I see is continuously updating inactive windows but in the background asynchronously. That being said, if we go with the current method, do you still think autocmd triggers for Feline updates should be user configurable?
> 
> I think asynchronous behaviour in neovim is hard to get right but if you do have a solution, I don't mind reviewing it.
> 
> I don't think there is any harm in making autocmds user configurable, since if it breaks, it will be on the end user and their configuration. Having the option to extend the autocmds sounds nice.

It was very simple to implement async. Just had to remove the autocmd triggers for update_inactive_windows and did this:
```lua
vim.loop.new_timer():start(0, vim.o.updatetime, vim.schedule_wrap(M.update_inactive_windows))
```

Ideally we could make the timer configurable too

ram02z:
> 
> It was very simple to implement async. Just had to remove the autocmd triggers for update_inactive_windows and did this:
> ```lua
> vim.loop.new_timer():start(0, vim.o.updatetime, vim.schedule_wrap(M.update_inactive_windows))
> ```
> 
> Ideally we could make the timer configurable too

Oh that's what you mean, sorry I am tired, I was thinking of multithreading .If I recall correctly, leaving the timer handle open can cause memory leaks.

famiu:
> > It was very simple to implement async. Just had to remove the autocmd triggers for update_inactive_windows and did this:
> > ```lua
> > vim.loop.new_timer():start(0, vim.o.updatetime, vim.schedule_wrap(M.update_inactive_windows))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Ideally we could make the timer configurable too
> 
> Oh that's what you mean, sorry I am tired

It's fine

> I was thinking of multithreading

I guess I could give that a shot too

> If I recall correctly, leaving the timer handle open can cause memory leaks

And how would that be fixed?

ram02z:
> 
> > I was thinking of multithreading
> 
> I guess I could give that a shot too

I am not even sure if it is possible because neovim runs on a single thread.

> 
> > If I recall correctly, leaving the timer handle open can cause memory leaks
> 
> And how would that be fixed?
> 

I think you have to close the handler once you are done with it, but thinking more about it, I'm not sure if you would need to do that for a statusline plugin.

famiu:
> > > I was thinking of multithreading
> > 
> > 
> > I guess I could give that a shot too
> 
> I am not even sure if it is possible because neovim runs on a single thread.
> 

I mean, `vim.loop.new_thread` exists, so let me look into whether that can solve our problems or not.

> > > If I recall correctly, leaving the timer handle open can cause memory leaks
> > 
> > 
> > And how would that be fixed?
> 
> I think you have to close the handler once you are done with it, but thinking more about it, I'm not sure if you would need to do that for a statusline plugin.

Precisely. When are we "done with" the handler?

ram02z:
I think the async code you posted is pretty much what the CursorHold autocmd already does. I think maybe just using `vim.schedule` instead of the one shot timer function, `vim.defer_fn`, would be more reliable, even though my testing had similar results.

famiu:
> I think the async code you posted is pretty much what the CursorHold autocmd already does.

Except it does it asynchronously so it doesn't block Neovim.

> I think maybe just using `vim.schedule` instead of the one shot timer function, `vim.defer_fn`, would be more reliable, even though my testing had similar results.

It does seem more reliable

famiu:
We could have the best of both worlds and asynchronously run `update_inactive_windows` everytime an autocmd trigger happens. But the simplest way to do that would be through plenary's async lib, and I don't know if that'd be worth adding an extra dependency.

famiu:
So I've rebased this to the latest master and also added an `update_triggers` configuration option. This seems to be ready for merge now unless we go the async route

famiu:
That being said, I think I'm going to merge this now, we've waited too long on this. If the merge causes any problems, please open a new issue.

xeluxee:
I confirm this works really well!
Thanks @famiu @ram02z

PR #56, opened 2021-09-08
feat: make components use window number

famiu:
This needs to be merged in order to make further progress on #55.

I changed quite a few things, so I expect a lot of breakages. I'd really love if people take the time to test it out before I merge this.

### DIFF
--- a/lua/feline/generator.lua
+++ b/lua/feline/generator.lua
@@ -171,12 +171,12 @@ local function parse_icon(icon, parent_hl)
 end
 
 -- Parse component provider
-local function parse_provider(provider, component)
+local function parse_provider(provider, component, winnr)
     local icon
     if type(provider) == "string" and type(providers[provider]) == "function" then
-        provider, icon = providers[provider](component)
+        provider, icon = providers[provider](component, winnr)
     elseif type(provider) == "function" then
-        provider, icon = provider(component)
+        provider, icon = provider(component, winnr)
     end
 
     if type(provider) ~= "string" then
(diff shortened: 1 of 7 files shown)


lua/feline/generator.lua
```diff
@@ -230,25 +229,25 @@ local function parse_statusline_section_old(section, type)
...
+function M.generate_statusline(winnr)
     if not M.components then
-        return ""
+        return ''
     end
 
     local statusline_type
 
-    if is_active and not is_forced_inactive() then
+    if winnr == vim.api.nvim_get_current_win() and not is_forced_inactive() then
```
ram02z, line 250: This will always evaluate to false since winnr is nil

lua/feline/providers/lsp.lua
```diff
@@ -1,59 +1,67 @@
 local M = {}
 
-function M.is_lsp_attached()
-    return next(vim.lsp.buf_get_clients()) ~= nil
+local lsp = vim.lsp
```
ram02z, line 3: https://github.com/famiu/feline.nvim/pull/33 was merged to stop loading it eagerly.

famiu: Oops, forgot about that. Fixed now

lua/feline/generator.lua
famiu: Added default value for winnr to work around that for now. Once this is merged I'll make proper use and actually pass the window number to the statusline generator

famiu:
If this PR is merged by itself, it will leave the `master` branch in a very buggy, unusable state. This PR needs the changes in #55 to function properly. So I'll be closing this and incorporating the changes in this PR to #55.

PR #58, opened 2021-09-09
Update all windows using vim.api.nvim_win_call

### LINKED ISSUE #52: [Bug] statusline is updated only in current window

xeluxee:
Alternative to #55. Fixes #52.
It works for me. 
@famiu @ram02z can you review it?

### DIFF
--- a/lua/feline/init.lua
+++ b/lua/feline/init.lua
@@ -9,12 +9,14 @@ local M = {}
 M.reset_highlights = gen.reset_highlights
 
 function M.update_all_windows()
-    for _, winid in ipairs(api.nvim_list_wins()) do
-        vim.wo[winid].statusline = api.nvim_win_call(winid,
-            function()
-                return require('feline').statusline(winid)
-            end
-        )
+    for _, winid in ipairs(api.nvim_tabpage_list_wins(0)) do
+        if api.nvim_win_get_config(winid).relative == '' then
+            vim.wo[winid].statusline = api.nvim_win_call(winid,
+                function()
+                    return require('feline').statusline(winid)
+                end
+            )
+        end
     end
 
     -- Reset local statusline of current window to use the global statusline for it
@@ -129,8 +131,9 @@ function M.setup(config)
     }, 'feline')
 end
 
-function M.statusline()
-    if g.statusline_winid == fn.win_getid() then
+function M.statusline(winid)
+    winid = winid or fn.win_getid()
+    if g.statusline_winid == winid then
         return gen.generate_statusline(true)
     else
         return gen.generate_statusline(false)
--- a/lua/feline/init.lua
+++ b/lua/feline/init.lua
@@ -8,6 +8,19 @@ local M = {}
 
 M.reset_highlights = gen.reset_highlights
 
+function M.update_all_windows()
+    for _, winid in ipairs(api.nvim_list_wins()) do
+        vim.wo[winid].statusline = api.nvim_win_call(winid,
+            function()
+                return require('feline').statusline(winid)
+            end
+        )
+    end
+
+    -- Reset local statusline of current window to use the global statusline for it
+    vim.wo.statusline = nil
+end
+
 local function parse_config(config_dict, config_name, expected_type, default_value)
     if config_dict and config_dict[config_name] then
         if type(config_dict[config_name]) == expected_type then
@@ -103,9 +116,16 @@ function M.setup(config)
     vim.o.statusline = '%!v:lua.require\'feline\'.statusline()'
 
     create_augroup({
-        {'WinEnter,BufEnter', '*', 'set statusline<'},
-        {'WinLeave,BufLeave', '*', 'lua vim.wo.statusline=require\'feline\'.statusline()'},
-        {'ColorScheme', '*', 'lua require\'feline\'.reset_highlights()'}
+        {
+            'WinEnter,BufEnter,WinLeave,BufLeave,SessionLoadPost,FileChangedShellPost',
+            '*',
+            'lua require("feline").update_all_windows()'
+        },
+        {
+            'SessionLoadPost,ColorScheme',
+            '*',
+            'lua require("feline").reset_highlights()'
+        }
     }, 'feline')
 end
 


lua/feline/init.lua
```diff
@@ -8,6 +8,19 @@ local M = {}
 
 M.reset_highlights = gen.reset_highlights
 
+function M.update_all_windows()
+    for _, winid in ipairs(api.nvim_list_wins()) do
+        vim.wo[winid].statusline = api.nvim_win_call(winid,
+            function()
+                return require('feline').statusline(winid)
```
ram02z, line 15: No need to pass winid

lua/feline/init.lua
```diff
@@ -8,6 +8,19 @@ local M = {}
 
 M.reset_highlights = gen.reset_highlights
 
+function M.update_all_windows()
+    for _, winid in ipairs(api.nvim_list_wins()) do
```
ram02z, line 12: Wrap the inside of the for loop, so it skips updating if the window is relative (floating).
i.e. ` if api.nvim_win_get_config(winid).relative == '' then ... end`

ram02z (commented):
Elegant solution, works well so far. I have tested it for your use case and using the general config to check for any issues.

lua/feline/init.lua
ram02z: Actually, it might be better to add winid as a parameter to statusline and use that to compare to `g.statusline_winid`

lua/feline/init.lua
ram02z: Or you could filter them
i.e. `local windows = vim.tbl_filter(is_regular_window, api.nvim_tabpage_list_wins(0))`
where `regular_window` is a function that chck if the winid is relative.

I think maybe using `nvim_tabpage_list_wins` is better since we only care about the windows in the current tabpage, therefore reducing unnecessary computation.

xeluxee:
@ram02z I implemented your suggestions.

But there is another problem, when LSP hover is shown current window loose focus and when hover floating window is closed current window doesn't gain focus again

ram02z:
> @ram02z I implemented your suggestions.
> 
> But there is another problem, when LSP hover is shown current window loose focus and when hover floating window is closed current window doesn't gain focus again, causing feline to be inactive even on current window.

Yeah, unfortunately checking the relative key doesn't stop the floating window from changing the current statusline.

There are two solutions in mind: 
1. Opening a floating window doesn't trigger the inactive state of the current non floating window.
2. Force a statusline update when a floating window is closed.

Not sure which is better to implement though.

lua/feline/init.lua
```diff
@@ -103,14 +118,22 @@ function M.setup(config)
     vim.o.statusline = '%!v:lua.require\'feline\'.statusline()'
 
     create_augroup({
-        {'WinEnter,BufEnter', '*', 'set statusline<'},
-        {'WinLeave,BufLeave', '*', 'lua vim.wo.statusline=require\'feline\'.statusline()'},
-        {'ColorScheme', '*', 'lua require\'feline\'.reset_highlights()'}
+        {
+            'WinEnter,BufEnter,WinLeave,BufLeave,SessionLoadPost,FileChangedShellPost',
```
ram02z, line 122: I tested this briefly and I think only WinEnter is required. It also seems to solve the issue @xeluxee was facing with the LSP hover window. This probably should require a little more testing due to the inactive state, but the statusline usually knows when to redraw themselves.

famiu: Having only WinEnter causes issues for me with vim-fugitive. I do have an alternative solution though. I kept the autocmds as is. How I instead fixed it is by not using `g.statusline_winid`, instead I made `update_all_windows` get the value of the current window and pass the value of is_active directly to Feline's statusline function. Like this:
```lua
function M.update_all_windows()
    local current_win = api.nvim_get_current_win()

    for _, winid in ipairs(api.nvim_tabpage_list_wins(0)) do
        if api.nvim_win_get_config(winid).relative == '' then

            vim.wo[winid].statusline = api.nvim_win_call(winid,
                function()
                    return require('feline').statusline(winid == current_win)
                end
            )
        end
    end

    -- Reset local statusline of current window to use the global statusline for it
    vim.wo.statusline = nil
end
```

Then the statusline function handles it accordingly:
```lua
function M.statusline(is_active)
    if is_active == nil then is_active = true end

    return gen.generate_statusline(is_active)
end
```

ram02z: > 
> Having only WinEnter causes issues for me with vim-fugitive. 

What issues are you facing? I think we should reduce the amount of autocmds, now that we update all windows in order to reduce redraws.

xeluxee: @ram02z removing some autocmds makes LSP hover working but breaks `NvimTreeToggle`, because when closing tree cursor remain in its window, so `WinEnter` isn't triggered

@famiu your suggestion doesn't work, when I open hover statusline becomes inactive

famiu: @xeluxee @ram02z removing only BufEnter and BufLeave seems to fix the issues

ram02z: >
> @ram02z removing some autocmds makes LSP hover working but breaks `NvimTreeToggle`, because when closing tree cursor remain in its window, so `WinEnter` isn't triggered

I installed nvim-tree, however, I cannot reproduce. Can you give me the steps to reproduce with `NvimTreeToggle`?

ram02z: > 
> 
> @xeluxee @ram02z removing only BufEnter and BufLeave seems to fix the issues. Though I don't know if that'd work with nvim-tree since I don't use it

Can you try if it works with only `WinEnter` and `WinLeave`? I think the other autocmds aren't required when updating all the windows.

xeluxee: @famiu removing Buf* seems to fix this issue, but causes the problem I described with NvimTree

@ram02z to reproduce my problem with NvimTree:
1. Open a buffer fullscreen
2. launch `:NvimTreeToggle`
3. See this:

st

famiu: > > @xeluxee @ram02z removing only BufEnter and BufLeave seems to fix the issues. Though I don't know if that'd work with nvim-tree since I don't use it
> 
> Can you try if it works with only `WinEnter` and `WinLeave`? I think the other autocmds aren't required when updating all the windows.

Pretty sure it'd be necessary to update statusline for all windows after a session is loaded or when a file is changed outside of Vim.

ram02z: Still cannot reproduce. This is only with `WinEnter` and with the changes suggested above by @famiu 
image

The numbers correspond to the values returned by `vim.fn.winnr()`

ram02z: >
> Pretty sure it'd be necessary to update statusline for all windows after a session is loaded or when a file is changed outside of Vim.

Actually, it might be fine to keep those autocmds since they rarely fire. However, I tested a session with multiple splits and it loaded fine.

famiu: > > Pretty sure it'd be necessary to update statusline for all windows after a session is loaded or when a file is changed outside of Vim.
> 
> Actually, it might be fine to keep those autocmds since they rarely fire. However, I tested a session with multiple splits and it loaded fine.

I guess we can remove `SessionLoadPost` then, but I still think `FileChangedShellPost` is necessary

famiu: It seems removing `WinLeave` doesn't cause any issues either

ram02z: >
> I guess we can remove `SessionLoadPost` then, but I still think `FileChangedShellPost` is necessary. Though I think it's better to just keep both of them

It is fine to keep both of them, since I don't think they will be called reguarly anyway.

xeluxee: @ram02z If I use only WinEnter my problem with nvimtree remains. That's really strange, but the problem disappear when i `:set nosplitright`. I think you have `splitright` set

ram02z: >
> @ram02z If I use only WinEnter my problem with nvimtree remains. That's really strange, but the problem disappear when i `:set nosplitright`. I think you have `splitright` not set

You are right, I have `nosplitright` set. I can now reproduce the issue. I think the issue is that setting `splitright` moves your cursor to the new split as well. I will test if `WinEnter` fires or not.

famiu: > > @ram02z If I use only WinEnter my problem with nvimtree remains. That's really strange, but the problem disappear when i `:set nosplitright`. I think you have `splitright` not set
> 
> You are right, I have `nosplitright` set. I can now reproduce the issue. I think the issue is that setting `splitright` moves your cursor to the new split as well, but invoking `NvimTreeToggle` doesn't ? I cannot seem to reproduce when invoking over plugins that create right/left splits; undotree and symbols-outline.

I'm wondering how that could be fixed, then

xeluxee: We'd need autocmds like `WinOpened` and `WinClosed`...

ram02z: >
> I'm wondering how that could be fixed, then

I have no idea, even after looking at the source. But no other plugins that open splits are misbehaving

> 
> We'd need autocmds like `WinOpened` and `WinClosed`...

`WinOpened` doesn't exist. I tried both `WinNew` and `WinClosed` and it doesn't fix the nvim-tree issue.

ram02z: > 
> ```lua
> function M.update_all_windows()
>     local current_win = api.nvim_get_current_win()
> 
>     for _, winid in ipairs(api.nvim_tabpage_list_wins(0)) do
>         if api.nvim_win_get_config(winid).relative == '' then
> 
>             vim.wo[winid].statusline = api.nvim_win_call(winid,
>                 function()
>                     return require('feline').statusline(winid == current_win)
>                 end
>             )
>         end
>     end
> ```

I think we don't actually have to update all the windows, since we only care about updating the inactive ones. Therefore I believe, `update_all_windows` should be change to `update_inactive_windows`. We can use @famiu 's snippet and make the following changes:

- ` if api.nvim_win_get_config(winid).relative == '' and winid ~= current_win then` 
- `return require('feline').statusline(false)`

I have tested it briefly and everything is working so far. I will continue to use this PR with this patch, https://gist.github.com/ram02z/39aff6de9a54c8a45bfe9035b5cf4183, and will let you know if I find any issues.

famiu: > @famiu your suggestion doesn't work, when I open hover statusline becomes inactive

Cannot reproduce the issue, works perfectly fine for me.

ram02z: Ok, so `VimEnter` or `BufReadPost` are also needed. I use a plugin called `committia.vim` which creates three splits when calling `git commit` from the shell. I think `VimEnter` would be best suited.

famiu: > Ok, so `VimEnter` or `BufReadPost` are also needed. I use a plugin called `committia.vim` which creates three splits when calling `git commit` from the shell and inactive splits aren't processed. I think `VimEnter` would be best suited.

But if we add it for VimEnter, it might still cause issues with other plugins that might create multiple splits, right? I think what we should do is add an autocmd for WinNew

famiu: I also realized that using this method, it's impossible for providers to get the actual current window and act on it, which might be crucial for some providers.

ram02z: >
> But if we add it for VimEnter, it might still cause issues with other plugins that might create multiple splits, right? I think what we should do is add an autocmd for WinNew.

I can try WinNew, but the VimEnter autocmd would only be limited to cases where the splits are created when nvim is started from shell anyway.

> 
> I also realized, if you close an inactive window, the window numbers probably wouldn't update on the other statusline (can't check it right now so I'm not sure but I think that's the case)

The only case I have tested was two plugins that open seperate splits (undotree and symbols-outline) and toggling either one of them still triggered `WinEnter`.  I think we should definitely hold off on merging this, until we can finalise the autocmds the would be needed to update all inactive windows.

famiu: > I also realized that using this method, it's impossible for providers to get the actual current window and act on it, which might be crucial for some providers, whereas #55 would allow to get access to both the window that the statusline is being generated for (through the `winnr` parameter), and the current window (through `nvim_get_current_win`)

@ram02z thoughts on this?

ram02z: > 
> 
> I also realized that using this method, it's impossible for providers to get the actual current window and act on it, which might be crucial for some providers, whereas #55 would allow to get access to both the window that the statusline is being generated for (through the `winnr` parameter), and the current window (through `nvim_get_current_win`)

Yeah, that is a limititation of this PR, however, I think adding `winnr` as a named parameter should probably be a seperate PR. I prefer the approach this PR is taking in order to update inactive windows. Honestly, adding `winnr` as a named parameter should also be given more consideration as it does increase the complexity of the codebase and I am not sure of how useful it can be.

ram02z: > 
>
> @ram02z thoughts on this? I think this approach is fundamentally a bit hacky and #55, although more complicated, is just more a lot reliable and a lot less like throwing darts in the dark

I can see why you may think its like throwing darts in the dark, but I believe using too many autocmds to redraw all the statuslines on the screen is pretty inefficient and can cause issues like we had with `BufEnter` and lsp hover. I do think having `winnr` as a named parameter is more reliable though.

famiu: > Honestly, adding `winnr` as a named parameter should also be given more consideration as it does increase the complexity of the codebase and I am not sure of how useful it can be.

@ram02z Honestly, at this point it's a tradeoff between simplicity and reliability. This PR is more simple but has proven to be quite unpredictable. #55 adds complexity to the codebase but in exchange provides much higher reliability over this PR. Not to mention this PR also a bit hacky and doesn't allow providers to access the actual current window, which a lot of providers may need to do. And if we want to provide that, we'd have to add support for a winnr paramter which'd add the same amount of complexity to the codebase.

Considering all that, I propose this:
We go with #55 but instead of `winnr`, we allow components to take a `ctx` parameter. As of now that'd only contain the window number (and possible buffer number too since a lot of providers need it), but if needed in the future, we could add additional details to it so the codebase doesn't get more complicated in the future.

ram02z: @famiu THe unpredictablity is an issue, so therefore I think we should continue with #55. I have moved the discussion there and would like to hear your opinions on my comment https://github.com/famiu/feline.nvim/pull/55#issuecomment-916792231

famiu:
Closing this since #55 was merged

PR #64, opened 2021-09-11
fix: don't defer current window reset

### LINKED ISSUE #63: [Bug] Feline occasionally shows inactive statusline on active window
Feline seems to occasionally show the inactive statusline on the active window. It seems to happen randomly and I didn't find a pattern to it yet, but it only happens when I switch tabs through mouseclick using [barbar.nvim](https://github.com/romgrk/barbar.nvim). Need to find the cause of this.

ram02z:
Attempt to fix #63. Can you test this @famiu?

### DIFF
--- a/lua/feline/init.lua
+++ b/lua/feline/init.lua
@@ -57,10 +57,10 @@ function M.update_inactive_windows()
                 vim.wo[winid].statusline = M.statusline(winid)
             end
         end
-
-        -- Reset local statusline of current window to use the global statusline for it
-        vim.wo.statusline = nil
     end)
+
+    -- Reset local statusline of current window to use the global statusline for it
+    vim.wo.statusline = nil
 end
 
 function M.setup(config)


famiu:
The problem is that #63 happens randomly so it's hard to reliably reproduce. But I've been clicking around tabs in barbar.nvim for a while and the issue didn't occur with the changes in this PR. So I guess I'll merge this since the change is harmless anyway, and I'll refer back to this if the issue ever does return.

ram02z:
Unforunately, the issue with using `vim.schedule` is that we aren't using it for its designed use case, it was meant to schedule a callback for the execution on the main event loop, whereas we are using it to defer the autocmd to make it execute last, so it is a bit of a hack. We could use `vim.defer_fn` but the problem is that I didn't notice a difference between using `vim.defer_fn` and `vim.schedule` anyway (that is largely due to my lack of knowledge of neovim's event loop).

famiu:
> Unforunately, the issue with using `vim.schedule` is that we aren't using it for its designed use case, it was meant to schedule a callback for the execution on the main event loop, whereas we are using it to defer the autocmd to make it execute last, so it is a bit of a hack. We could use `vim.defer_fn` but the problem is that I didn't notice a difference between using `vim.defer_fn` and `vim.schedule` anyway (that is largely due to my lack of knowledge of neovim's event loop).

Considering that our only other option was to use a timer to update the statusline in inactive windows continuously, I wouldn't call this a particularly bad solution.

PR #65, opened 2021-09-11
feat: add `max_length` component option

### LINKED ISSUE #32: [Feature request]  custom truncated the long components

Unreasonable rendering of too long components.

For example:
```lua
-- config:
local cur_function = {
    provider = function()
        return vim.fn["nvim_treesitter#statusline"](90)
    end,
    enabled = function()
        return vim.NIL ~= vim.fn["nvim_treesitter#statusline"](90)
    end,
    left_sep = ' ',
    hl = {
        fg = colors.yellow
    }
}

local components = {
    left = {
        active = {
            comps.vi_mode.left,
            comps.file.info,
            cur_function,
-- .......
        },
        inactive = {
            comps.file.info
        }
    },
-- .........
}

require'feline'.setup {
-- .....
    components = components,
-- ....
}

```

it work like that:
截屏2021-08-20 上午10 20 11

but if the component too long , it work like that:
截屏2021-08-20 上午10 31 54

`comps.vi_mode.left` ,  `comps.file.info`  was truncated.

and now i solve it by a status line mark `%<`
```
provider = function()
    return "%<" .. vim.fn["nvim_treesitter#statusline"](90)
end
```

it work like that:
截屏2021-08-20 上午10 56 33

However, display function name is more usefully than function signature

A clear and concise description of what you want to happen. I hope it can truncate.

Therefore,  i hope we can cut off the tail rather than the head of component, like that :

截屏2021-08-20 上午11 06 53

Easy goal: cut off the tail for some component.

Future hope:
It just an idea without detail.
I hope each component can set 3  properties: `max_size`, `min_size`, `truncate_level`. For the too long component, it can truncate the low level component to min_size. 

**Additional context**

famiu:
Closes #32

Added support for component truncation through a `max_length` component option. Implementation is quite simple so I don't see a way for it to go wrong. Would still like to test it for a bit before merging it, however. Whether it even needs to be a feature at all is also up for debate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,6 +323,8 @@ hl = function()
 end
 ```
 
+- `max_length` (number): Maximum length of the component. If the component exceeds this length in the statusline, it'll be truncated with a `...` at the end to signify the truncation. `max_length` is not set to any component by default, so by default components don't have any limit to how large they can be. Note that since a `...` is appended at the end of a truncated component, the value of `max_length` must always be larger than 3, otherwise it'll be ignored by Feline.
+
 <br>
 
 **NOTE:** Some providers may also have special component values unique to them, such as the `file_info` provider having a `file_modified_icon` value that you can set. For more info, see: [default providers](#default-providers).
@@ -433,7 +435,7 @@ components.active[2][1] = {
     provider = 'file_encoding'
 }
 
--- Component that shows current git branch
+-- Component that shows current git branch with a maximum length of 8
 components.active[3][1] = {
     provider = 'git_branch',
     hl = {
@@ -441,6 +443,7 @@ components.active[3][1] = {
         bg = 'black',
         style = 'bold'
     },
+    max_length = 8,
     right_sep = function()
         local val = {hl = {fg = 'NONE', bg = 'black'}}
         if vim.b.gitsigns_status_dict then val.str = ' ' else val.str = '' end
--- a/lua/feline/generator.lua
+++ b/lua/feline/generator.lua
@@ -189,6 +189,10 @@ local function parse_component(component, winid)
 
     icon = parse_icon(evaluate_if_function(icon), hl)
 
+    if component.max_length and component.max_length > 3 and #str > component.max_length then
+        str = str:sub(1, component.max_length - 3) .. '...'
+    end
+
     return left_sep_str .. icon .. '%#' .. hlname .. '#' .. str .. right_sep_str
 end
 


ram02z:
I think the functionality is basic enough to be opt-in, and like you said it shouldn't cause any issue. However, I think truncation has potiential to be a standard feature of feline. You would add it as a component value like `hl`. An example feature could be the ability to customise the truncation symbol instead of the default, `...`. However, for it become useful, it needs to be integrated into the provider so we can do things like check the length of the provider and check the remaining screen width to decide whether to truncate or not. An optimistic goal for this idea would allow for levels of truncation whereby the user could specify multiple return values to ensure that the truncation retains the information the user finds valuable, rather than just removing X trailing characters.

famiu:
> However, for it become useful, it needs to be integrated into the provider so we can do things like check the length of the provider and check the remaining screen width to decide whether to truncate or not.

That sounds good in theory but it sounds like it'd be a nightmare to actually implement, since providers would then have to interact with eachother and find an "optimal truncation method".

> An optimistic goal for this idea would allow for levels of truncation whereby the user could specify multiple return values to ensure that the truncation retains the information the user finds valuable, rather than just removing X trailing characters.

I don't exactly understand what you mean by multiple return values

ram02z:
> 
> That sounds good in theory but it sounds like it'd be a nightmare to actually implement, since providers would then have to interact with eachother and find an "optimal truncation method".

Yeah, I am just dumping my thoughts, and I think this a feature that will require an elaborate design process for it work as intended.

> 
> I don't exactly understand what you mean by multiple return values

Lets use the example shown in #32. When no truncation is required, we will return the full function signature. However, when space on the statusline becomes scarce, we might not want half of the function signature since its incomplete, instead we might show the function name instead.

famiu:
> > That sounds good in theory but it sounds like it'd be a nightmare to actually implement, since providers would then have to interact with eachother and find an "optimal truncation method".
> 
> Yeah, I am just dumping my thoughts, and I think this a feature that will require an elaborate design process for it work as intended.

Yes, that makes me wonder if it'd be worth all the hassle implementing it.

> > I don't exactly understand what you mean by multiple return values
> 
> Lets use the example shown in #32. When no truncation is required, we will return the full function signature. However, when space on the statusline becomes scarce, we might not want half of the function signature since its incomplete, instead we might show the function name instead.

I mean, things like that will be needed to be created by the user themselves for custom providers. The best we can do is add a default truncation to the built-in providers.

ram02z:
> 
> I mean, things like that will be needed to be created by the user themselves for custom providers. The best we can do is add a default truncation to the built-in providers.

Yeah, currently I am doing it for that for myself as well, but it could a feature in the backlog. However, for the meanwhile, I think the `max_length` component should be implemented as a part of component, called `truncate` or the likes. That way it could also take an optional paramter `str`, which allows the user to change the the truncation ending or even set it to an empty string to remove it entirely.

famiu:
> > I mean, things like that will be needed to be created by the user themselves for custom providers. The best we can do is add a default truncation to the built-in providers.
> 
> Yeah, currently I am doing it for that for myself as well, but it could a feature in the backlog. However, for the meanwhile, I think the `max_length` component should be implemented as a part of component, called `truncate` or the likes. That way it could also take an optional paramter `str`, which allows the user to change the the truncation ending or even set it to an empty string to remove it entirely.

I still have doubts on whether that would be worth it. But I'll look into whether I can work on it. In case I do implement it, I'm wondering if I should modify this PR close this PR and create a new one since the truncation is a lot more robust than just adding a `max_length` component value

ram02z:
> 
> I still have doubts on whether that would be worth it. But I'll look into whether I can work on it. In case I do implement it, I'm wondering if I should modify this PR close this PR and create a new one since the truncation is a lot more robust than just adding a `max_length` component value

I mean the only component values that would be useful for truncation personally are `max_length` and `str`, so I think modifying this PR might be fine. Maybe someone may find it useful to truncate from the front of the string instead, but I don't see a use case for that.

famiu:
> > I still have doubts on whether that would be worth it. But I'll look into whether I can work on it. In case I do implement it, I'm wondering if I should modify this PR close this PR and create a new one since the truncation is a lot more robust than just adding a `max_length` component value
> 
> I mean the only component values that would be useful for truncation personally are `max_length` and `str`, so I think modifying this PR might be fine. Maybe someone may find it useful to truncate from the front of the string instead, but I don't see a use case for that.

So I've thought of a possible implementation for truncation. It goes like this:

1. The truncation property, along with`str`, will have a `priority` value, which denotes the priority of the provider. The higher the priority, the less likely it is to be truncated.

2. `parse_component` will return the priority of the provider alongside the parsed component. It will no longer return a string as the parsed component, but rather a table containing the parsed left_sep, the parsed right_sep, the highlight name, the icon and the component string.

3. There will be a new `truncate_components` function which will be called by `generate statusline` that takes a table of all the components and their priorities. The function would find the screen width and truncate the component strings ordered by priority in ascending order (and if multiple components have the same priority, the longer ones would be truncated first) until the length of the statusline is lesser than or equal to the size of the window, and then the statusline will be shown.

4. Each of the built-in providers could have special built-in truncation functions that would be used by `truncate_components` instead of the default truncation method.

5. We'd somehow have to trigger an update of inactive windows when windows are resized (I'm not aware how to do that since there's no WinResized autocmd).

Now there's 5 major issues I see with this:

1. This will be extremely difficult to implement, probably not worth the hassle.

2. The complexity of the codebase would be greatly increased with little payoff.

3. It will also have a pretty big performance overhead, even for those who don't wish to use it.

4. Even if we do decide to go with it, it's still hard to choose where and how much to truncate each component.

5. Truncation can already be implemented by the user themselves, and honestly when it comes to things like truncation, I don't think there's a one-size-fits-all solution. Since it would be different for each user, I think it should be implemented by the user themselves.

All that being said, I'd say we drop this idea, and this PR as well since `max_length` is a very simple implementation and not very useful either honestly thinking about it.

ram02z:
> 
>  The truncation property, along with`str`, will have a `priority` value, which denotes the priority of the provider. 
> The higher the priority, the less likely it is to be truncated.

I had this idea for my own config and I agree it is easier to implement in someone's own config than directly in feline. The rest of the points you bring up are also valid. Personally, I wouldn't use the `max_length` option since I can just truncate my components through the provider directly.

 I could create a PR to add a README section that discusses how you can truncate components. It might even be time to add a wiki for user contributed tips.

famiu:
> > The truncation property, along with`str`, will have a `priority` value, which denotes the priority of the provider.
> > The higher the priority, the less likely it is to be truncated.
> 
> I had this idea for my own config and I agree it is easier to implement in someone's own config than directly in feline. The rest of the points you bring up are also valid. Personally, I wouldn't use the `max_length` option since I can just truncate my components through the provider directly.

Which is why I plan to drop this PR.

> I could create a PR to add a README section that discusses how you can truncate components. It might even be time to add a wiki for user contributed tips.

I could make a wiki. It'd also allow putting how to configure the statusline in the wiki since the README is rather large, would like to hear your thoughts on that.

As of now, I think you should make a PR into the README first so I can later use it for the wiki in case I decide to put the other parts of README into the wiki as well, and also because I'm not really experienced with GitHub wiki pages.

PR #68, opened 2021-09-12
fix: make disable inherit correct table

ram02z:
If you have a custom `force_inactive` table defined, it will instead disable the statusline for the respective files/buffers.


famiu:
Ah, that was a silly mistake on my part. Will merge the PR

Issue #71, opened 2021-09-15
[Feature] Hide separator if provider result is empty


b0o:
I have some components that are only displayed on certain conditions and I don't want their separators to show if the component provider returns the empty string.

Right now, I could call the provider function again inside the separator function or set a variable, but both feel messy and unnecessary. It would be nice if the separator table could have a key `omit_empty` which, if true, would cause the separator to be ignored in this case, e.g.:

```lua
config.components.active[2] = {
    {
        provider = 'lsp_clients_running',
        hl = { fg = 'green' },
        right_sep = { str = ' ', omit_empty = true },
    },
    {
        provider = 'lsp_clients_exited_ok',
        hl = { fg = 'lavender' },
        right_sep = { str = ' ', omit_empty = true },
    },
    {
        provider = 'lsp_clients_exited_err',
        hl = { fg = 'red' },
        right_sep = { str = ' ', omit_empty = true },
    },
    -- ...
}
```

My current solution (ugly, inefficient):
```lua
    {
        provider = 'lsp_clients_running',
        hl = { fg = 'green' },
        right_sep = function()
          return {
            str = user_lsp_status.clients_running() == '' and '' or ' '
          }
        end
    },
    {
        provider = 'lsp_clients_exited_ok',
        hl = { fg = 'lavender' },
        right_sep = function()
          return {
            str = user_lsp_status.clients_exited_ok() == '' and '' or ' '
          }
        end
    },
    {
        provider = 'lsp_clients_exited_err',
        hl = { fg = 'red' },
        right_sep = function()
          return {
            str = user_lsp_status.clients_exited_err() == '' and '' or ' '
          }
        end
    },
    -- ...
```

famiu:
I actually went one step further and made separators hidden by default if the component string is empty. And I added an `always_visible` option to make it visible.

b0o:
I like that even better! Thank you for such a quick response!

PR #72, opened 2021-09-15
feat: add omit_empty key to separators

ram02z:
Ref: #71

I was also using similar workarounds as the issue's author and think this a reasonable use case. Not sure about the name of the key though. Let me know if you have suggestions before I go ahead and add the proper documentation (function comment and README).

### DIFF
--- a/lua/feline/generator.lua
+++ b/lua/feline/generator.lua
@@ -108,7 +108,7 @@ end
 -- Parse component seperator
 -- By default, foreground color of separator is background color of parent
 -- and background color is set to default background color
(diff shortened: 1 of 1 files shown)


famiu:
> Ref: #71
> 
> I was also using similar workarounds as the issue's author and think this a reasonable use case. Not sure about the name of the key though. Let me know if you have suggestions before I go ahead and add the proper documentation (function comment and README).

I didn't notice that you made a PR and already ended up implementing this. I'm going to close this since it's already been implemented

PR #74, opened 2021-09-16
fix: ensure that parent_bg has a value

ram02z:
Closes: #23

`parent_bg` can be `nil`, which resulted in the the splash screen bug.

cc: @ibhagwan @famiu

### DIFF
--- a/lua/feline/generator.lua
+++ b/lua/feline/generator.lua
@@ -209,8 +209,8 @@ local function parse_component(component, winid)
 
     local is_component_empty = str == ''
 
-    local left_sep_str = parse_sep_list(component.left_sep, hl.bg, is_component_empty)
-    local right_sep_str = parse_sep_list(component.right_sep, hl.bg, is_component_empty)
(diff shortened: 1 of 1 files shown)


famiu:
> Closes: #23
> 
> `parent_bg` can be `nil`, which resulted in the the splash screen bug.
> 
> cc: @ibhagwan @famiu

Can you base your changes off of the develop branch instead, since master is supposed to only be reserved for final releases now

famiu:
For some reason, the PR is showing 2 commits instead of one, it'd be nice if you could fix that, you probably need to rebase your changes off of develop

ram02z:
Not sure why it does that on github. I can't rebase because it is up to date.

famiu:
It seems it's fixed now

famiu:
@ram02z wouldn't it be better to add the check on `parse_sep_list` so it only happens once instead of multiple times for a separator list?

ram02z:
Turns out I added a colon after "Closes" in the PR description, so it didn't close the issue automatically :/

ibhagwan:
Fantastic fix @ram02z!

PR #76, opened 2021-09-16
feat: add auto vimdoc generation workflow

ram02z:
I saw a comment on the reddit post you made about the lack of a vimdoc file. I have written docs for people in the past but this README is massive (that's a good thing), so I took inspiration from shadmansaleh's lualine and tried out https://github.com/kdheepak/panvimdoc. 

The github workflow is mostly borrowed from https://github.com/shadmansaleh/lualine.nvim/blob/master/.github/workflows/docs.yml. 

#### Changes
- The workflow runs whenever there is a change in the README pushed to any branch, this should probably be limited to the develop branch, however, I wanted your input first. 
- I had to add an empty an empty doc file in this PR so the workflow can submit a PR. 
- Also, I changed the formatting slightly in the README to accomodate for the generated vimdoc.

#### Issues
- Not sure if the `HOW TO INSTALL` section should be ignored, but I think it should since users would have already installed it.
- Some of the README headings result in odd vimdoc links like `*feline-The-setup-function*`.

The current generated vimdoc can be seen here: https://github.com/ram02z/feline.nvim/pull/11

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,16 @@
+<!-- panvimdoc-ignore-start -->
+
 # feline.nvim
 
 A minimal, stylish and customizable statusline for Neovim written in Lua
 
 Requires Neovim >= 0.5
 
+
+<!-- panvimdoc-ignore-end -->
 ## About
 
-Feline is a Lua statusline that prioritizes speed, customizability and minimalism. It's blazing fast and never gets in your way. Feline only provides you with the necessary tools that you need to customize the statusline to your liking and avoids feature-bloat. It's also extremely customizable and allows you to configure it in any way you wish to. Feline also has reasonable defaults for those who don't want to configure things and just want a good out of the box experience. 
+Feline is a Lua statusline that prioritizes speed, customizability and minimalism. It's blazing fast and never gets in your way. Feline only provides you with the necessary tools that you need to customize the statusline to your liking and avoids feature-bloat. It's also extremely customizable and allows you to configure it in any way you wish to. Feline also has reasonable defaults for those who don't want to configure things and just want a good out of the box experience.
 
 ## Features
 
@@ -33,6 +37,8 @@ Feline is a Lua statusline that prioritizes speed, customizability and minimalis
   - [nvim-web-devicons](https://github.com/kyazdani42/nvim-web-devicons/) - For icon support
   - [A patched font](https://github.com/ryanoasis/nerd-fonts/) - For icon support
   - [gitsigns.nvim](https://github.com/lewis6991/gitsigns.nvim/) - For [git providers](#git)
+
+<!-- panvimdoc-ignore-start -->
 
 ## Screenshots
 
@@ -75,7 +81,7 @@ Feline is a Lua statusline that prioritizes speed, customizability and minimalis
 
 You can add your own configuration to this list. If you're interested, simply make a [Pull Request](CONTRIBUTING.md) and I'll add it.
 
-## How to install
+## Installation
 
 ### Install latest stable release
 
@@ -109,7 +115,9 @@ Plug 'famiu/feline.nvim', { 'tag': 'v0.1' }
 
 ### Install from develop branch
 
-**NOTE:** The develop branch is intended for those who either wish to contribute to Feline or want the absolute bleeding edge version of Feline. The develop branch may contain changes that have not been tested thoroughly and therefore stability is not guaranteed. Moreover, commit history of the develop branch may be rewritten at any time. So use it at your own risk.
+**NOTE:** The develop branch is intended for those who either wish to contribute to Feline or want the absolute bleeding edge version of Feline.
+The develop branch may contain changes that have not been tested thoroughly and therefore stability is not guaranteed.
+Moreover, commit history of the develop branch may be rewritten at any time. So use it at your own risk.
 
 - [packer.nvim](https://github.com/wbthomason/packer.nvim/):
 
@@ -123,7 +131,9 @@ use { 'famiu/feline.nvim', branch = 'develop' }
 Plug 'famiu/feline.nvim', { 'branch': 'develop' }
 ```
 
-## How to use
+<!-- panvimdoc-ignore-end -->
+
+## Usage
 
 Once you've installed Feline, it's very easy to start using it. Here are the following options for using Feline:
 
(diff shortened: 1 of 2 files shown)


famiu:
Haven't looked into the files yet (will do soon) but I think it should ignore the How To Install section

famiu:
Why are the headers in "Tips and tricks" and "Common issues" changed to quotes

famiu:
I'm not experienced with Vim docs so I could be wrong here, but shouldn't I be able to do `:help feline` after putting the generated docs as `doc/feline.txt`? Currently I'm not able to do that

shadmansaleh:
> but shouldn't I be able to do :help feline after putting the generated docs as doc/feline.txt? 

Run `:helptags doc/` to generate tags . Plugin managers will run it on userside

famiu:
> > but shouldn't I be able to do :help feline after putting the generated docs as doc/feline.txt?
> 
> Run `:helptags doc/` to generate tags . Plugin managers will run it on userside
> 
> Also do yourself a favor add `/doc/tags` to gitignore.

Ah alright, will keep that in mind

famiu:
I made some changes to README.md, you should probably rebase your changes against the latest commit

ram02z:
> 
> 
> Why are the headers in "Tips and tricks" and "Common issues" changed to quotes

The headings resulted in auto generated vimdoc headings like `feline-Feline-crashes-or-disappears-for-seemingly-no-reason`

ram02z:
@famiu I changed the `How to use` heading to `Usage` and limited the workflow to the develop branch. I also ignored the `How to install` section.

famiu:
> > Why are the headers in "Tips and tricks" and "Common issues" changed to quotes
> 
> The headings resulted in auto generated vimdoc headings like `feline-Feline-crashes-or-disappears-for-seemingly-no-reason`

but that isn't a problem for the "Tips and tricks" section, is it? Keep the quotes for "Common issues" I guess but I think "Tips and tricks" should have headers

ram02z:
> 
> but that isn't a problem for the "Tips and tricks" section, is it? Keep the quotes for "Common issues" I guess but I think "Tips and tricks" should have headers

Both "Common issues" and "Tips and tricks" have headings in the vimdoc. 

See https://github.com/ram02z/feline.nvim/blob/0d944ee7a675531b2d4d549a72effb08de098814/doc/feline.txt#L868-L878

famiu:
> > but that isn't a problem for the "Tips and tricks" section, is it? Keep the quotes for "Common issues" I guess but I think "Tips and tricks" should have headers
> 
> Both "Common issues" and "Tips and tricks" have headings in the vimdoc.
> 
> See https://github.com/ram02z/feline.nvim/blob/0d944ee7a675531b2d4d549a72effb08de098814/doc/feline.txt#L868-L878

I mean the headings inside Tips and tricks (eg: "Reset highlights") should also be headings instead of quotes

ram02z:
> 
> I mean the headings inside Tips and tricks (eg: "Reset highlights") should also be headings instead of quotes

Oh, I understand now. But that would result in a heading for "Use thin line instead of the inactive statusline" to become `feline-Use-thin-line-instead-of-the-inactive-statusline`, which I think is too long.

famiu:
> > I mean the headings inside Tips and tricks (eg: "Reset highlights") should also be headings instead of quotes
> 
> Oh, I understand now. But that would result in the heading for "Use thin line instead of the inactive statusline" to become `feline-Use-thin-line-instead-of-the-inactive-statusline`, which I think is too long.

How about rename it to "Thin line for horizontal splits"

famiu:
One last thing, you should probably rename "The setup function" to just "Setting up Feline" or maybe even just "Setup". Also remember to change all references to the setup function header in the README as well

famiu:
oops, accidentally clicked on "Close with comment"

ram02z:
> 
> One last thing, you should probably rename "The setup function" to just "Setting up Feline" or maybe even just "Setup". Also remember to change all references to the setup function header in the README as well

Yeah, to follow along with the theme, I will also change `How to install` to `Installation`

famiu:
> > One last thing, you should probably rename "The setup function" to just "Setting up Feline" or maybe even just "Setup". Also remember to change all references to the setup function header in the README as well
> 
> Yeah, to follow along with the theme, I will also change `How to install` to `Installation`

Yeah, feel free to change any other header you feel is necessary to change as well.

ram02z:
> 
> 
> One last thing, you should probably rename "The setup function" to just "Setting up Feline" or maybe even just "Setup". Also remember to change all references to the setup function header in the README as well

I settled with `Setup function`, since `Setup` on it's own was abit ambigious. Should be ready to merge now.

ram02z:
Hmm, not sure why `the-setup-function` is still being referenced. 

See https://github.com/famiu/feline.nvim/blob/a15d0c990c9afb9ff6f50523222fa3a80e460003/doc/feline.txt#L669

famiu:
The help tags are also not appearing when I use the minimal_init. Isn't Packer supposed to setup the help tags?

README.md
```diff
@@ -510,7 +521,7 @@ hl = {bg = 'oceanblue'},
 right_sep = 'slant_right'
 ```
 
-Not only that, you can add your own custom colors and separators through [the setup function](#the-setup-function) which allows you to just use the name of the color or separator to refer to it.
+Not only that, you can add your own custom colors and separators through [the setup function](#setup-function) which allows you to just use the name of the color or separator to refer to it.
```
ram02z, line 524: I think I should have changed this to become `the [setup function](#setup-function)`

famiu: does changing it to that fix the issue?

ram02z:
> 
> 
> The help tags are also not appearing by default when I use the minimal_init. Isn't Packer supposed to setup the help tags?

My help tags generated after I compiled.

README.md
```diff
@@ -624,7 +635,7 @@ If you like the defaults for the most part but there's some things you want to c
 local components = require('feline.presets')[preset_name].components
 ```
 
-After that, you can just modify the components and call [the setup function](#the-setup-function) with the preset as you normally would.
+After that, you can just modify the components and call [the setup function](#setup-function) with the preset as you normally would.
```
famiu, line 638: Maybe you should change this to `the [setup function](#setup-function)` too

famiu:
> > The help tags are also not appearing by default when I use the minimal_init. Isn't Packer supposed to setup the help tags?
> 
> My help tags generated after I compiled.

What do you mean by "compiled" in this case?

ram02z:
> 
> What do you mean by "compiled" in this case?

After I ran `PackerCompile`, it showed up. You could also run `:helptags doc/` to force the generation of the tags.

famiu:
> > What do you mean by "compiled" in this case?
> 
> After I ran `PackerCompile`, it showed up. You could also run `:helptags doc/` to force the generation of the tags.

Ah. Alright, it seems to work fine now

PR #83, opened 2021-09-18
Fixes related to component highlight generation

### LINKED ISSUE #23: [Bug] feline causes the neovim splash screen to disappear

famiu:
~~Closes #23~~

Not sure if this is the most elegant solution, however, since it might cause some slowdowns for string values of `hl`

### DIFF
--- a/README.md
+++ b/README.md
@@ -777,7 +777,7 @@ end
 components.inactive = {
     {
         {
-            provider = '',
+            provider = ' ',
             hl = InactiveStatusHL
         }
     }
--- a/lua/feline/generator.lua
+++ b/lua/feline/generator.lua
@@ -258,9 +258,17 @@ local function parse_component(component, winid)
         icon = component.icon
     end
 
+    local hl_str
+
+    if not is_component_empty then
+        hl_str = string.format('%%#%s#', hlname)
+    else
+        hl_str = ''
+    end
+
     icon = parse_icon(evaluate_if_function(icon), hl)
 
-    return string.format('%s%s%%#%s#%s%s', left_sep_str, icon, hlname, str, right_sep_str)
+    return string.format('%s%s%s%s%s', left_sep_str, icon, hl_str, str, right_sep_str)
 end
 
 -- Parse components of a section of the statusline
--- a/lua/feline/generator.lua
+++ b/lua/feline/generator.lua
@@ -198,7 +198,6 @@ local function parse_icon(icon, parent_hl)
         str = icon
         hl = parent_hl
     else
-        icon = evaluate_if_function(icon)
         str = icon.str or ''
         hl = icon.hl or parent_hl
     end
@@ -251,7 +250,6 @@ local function parse_component(component, winid)
     -- If highlight is a table, just normally generate a highlight name
     else
         hl = parse_hl(hl)
-        hlname = get_hlname(hl)
     end
 
     local is_component_empty = str == ''
@@ -260,22 +258,21 @@ local function parse_component(component, winid)
     local right_sep_str = parse_sep_list(component.right_sep, hl.bg, is_component_empty)
 
     if is_component_empty then
-        icon = nil
-    elseif component.icon then
-        icon = component.icon
-    end
-
-    local hl_str
-
-    if not is_component_empty then
-        hl_str = string.format('%%#%s#', hlname)
+        return string.format(
+            '%s%s',
+            left_sep_str,
+            right_sep_str
+        )
     else
-        hl_str = ''
+        return string.format(
+            '%s%s%%#%s#%s%s',
+            left_sep_str,
+            parse_icon(evaluate_if_function(component.icon or icon), hl),
+            hlname or get_hlname(hl),
+            str,
+            right_sep_str
+        )
     end
-
-    icon = parse_icon(evaluate_if_function(icon), hl)
-
-    return string.format('%s%s%s%s%s', left_sep_str, icon, hl_str, str, right_sep_str)
 end
 
 -- Parse components of a section of the statusline
@@ -346,6 +343,8 @@ function M.generate_statusline(winid)
         end
     end
 
+    -- Never return an empty string since setting statusline to an empty string will make it
+    -- use the global statusline value (same as active statusline) for inactive windows
     if statusline_str == '' then
         return string.format('%%#%s#', defhl())
     else
--- a/lua/feline/generator.lua
+++ b/lua/feline/generator.lua
@@ -118,11 +118,18 @@ local function get_hlname(hl, parent_hl)
 
     hl = parse_hl(hl, parent_hl)
 
+    local fg_str, bg_str
+
+    -- If first character of the color starts with '#', remove the '#' and keep the rest
+    -- If it doesn't start with '#', do nothing
+    if hl.fg:sub(1, 1) == '#' then fg_str = hl.fg:sub(2) else fg_str = hl.fg end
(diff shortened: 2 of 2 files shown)


famiu:
Okay this seems to still have some issues, will be fixing them soon

famiu:
Okay, the issues seem to be dealt with now. What I'm worried about is any potential performance hits

famiu:
I did some benchmarking and performance does seem to be worse by about 10 to 20 microseconds compared to develop, let me see if I can improve that

famiu:
It suddenly doesn't work for me again anymore

famiu:
It seems this doesn't really close #23, but it still fixes a bug by actually getting the highlight values for string highlights since they're also used as parent_hl and parent_bg values

famiu:
Right now, the empty parts of the statusline use the highlight of the previous component due to how Vim statuslines work. While this can be seen as a bug that's fixed by adding `defhl()` at the end of every statusline string, it can also be seen as a feature which makes [this](https://github.com/famiu/feline.nvim#thin-line-for-horizontal-splits) work. I'm not sure whether to keep it or remove it. It can easily be fixed on user-side by just putting an empty component with their desired highlight as of now, but I'm not sure if that'd be the best decision to make. @ram02z thoughts?

ram02z:
> 
> 
> Right now, the empty parts of the statusline use the highlight of the previous component due to how Vim statuslines work. While this can be seen as a bug that's fixed by adding `defhl()` at the end of every statusline string, it can also be seen as a feature which makes [this](https://github.com/famiu/feline.nvim#thin-line-for-horizontal-splits) work. 

Hmm, not sure if the empty parts using the highlight of the previous component is a bug, since you did say that vim statuslines work like that.

>
> I'm not sure whether to keep it or remove it. It can easily be fixed on user-side by just putting an empty component with their > desired highlight as of now, but I'm not sure if that'd be the best decision to make. @ram02z thoughts?

Not sure what it exactly solves, but if it doesn't affect the redraw time then it should be fine to keep it. I am slightly confused, is this feature already implemented in feline or are you asking if you should implement it?

Btw, this PR does seem to resolve #23 for me. I tried the default config and my own config, which were both affected.

famiu:
> > Right now, the empty parts of the statusline use the highlight of the previous component due to how Vim statuslines work. While this can be seen as a bug that's fixed by adding `defhl()` at the end of every statusline string, it can also be seen as a feature which makes [this](https://github.com/famiu/feline.nvim#thin-line-for-horizontal-splits) work.
> 
> Hmm, not sure if the empty parts using the highlight of the previous component is a bug, since you did say that vim statuslines work like that.

True.

> > I'm not sure whether to keep it or remove it. It can easily be fixed on user-side by just putting an empty component with their > desired highlight as of now, but I'm not sure if that'd be the best decision to make. @ram02z thoughts?
> 
> Not sure what it exactly solves, but if it doesn't affect the redraw time then it should be fine to keep it. I am slightly confused, is this feature already implemented in feline or are you asking if you should implement it?

It already works like this, I'm just asking if it should continue to work like this or not, since it can sometimes give undesired results.

> Btw, this PR does seem to resolve #23 for me. I tried the default config and my own config, which were both affected.

Strange, this doesn't work for me, can you check again with the latest commit if it still works.

ram02z:
> 
> Strange, this doesn't work for me, can you check again with the latest commit if it still works.

Still works (neovim nightly).

famiu:
> > Strange, this doesn't work for me, can you check again with the latest commit if it still works.
> 
> Still works (neovim nightly).

I'm using Neovim 0.5 stable, maybe that's why it's still causing issues for me?

famiu:
To clarify further, for me it works with my config but not the default config

famiu:
After bisecting all the providers in the default preset, I've detected that the problem seems to come from the `git_branch` provider

famiu:
> > > Strange, this doesn't work for me, can you check again with the latest commit if it still works.
> > 
> > 
> > Still works (neovim nightly).
> 
> I'm using Neovim 0.5 stable, maybe that's why it's still causing issues for me?

switching to latest neovim doesn't fix my issues

famiu:
@ibhagwan can you try this pull and see if the splash screen works for you

famiu:
I'll be merging this now either way since this doesn't necessarily have to close #23, it already contains a lot of other fixes. Maybe I'll make another PR for #23

ibhagwan:
> @ibhagwan can you try this pull and see if the splash screen works for you

Works!

famiu:
> > @ibhagwan can you try this pull and see if the splash screen works for you
> 
> Works!

Strange.. why doesn't it appear for me

ibhagwan:
> > > @ibhagwan can you try this pull and see if the splash screen works for you
> > 
> > 
> > Works!
> 
> Strange.. why doesn't it appear for me

Just tested the merged `develop` branch, works there as well.

famiu:
> > > > @ibhagwan can you try this pull and see if the splash screen works for you
> > > 
> > > 
> > > Works!
> > 
> > 
> > Strange.. why doesn't it appear for me
> 
> Just tested the merged `develop` branch, works there as well.

can you try using the default preset with the develop branch and see if that works too

ibhagwan:
Works, on both 0.5 and the nightly:

screenshot-1632041455

famiu:
huh... then I wonder what's making it not work for me

ibhagwan:
> huh... then I wonder what's making it not work for me

Perhaps another plugin?

famiu:
> > huh... then I wonder what's making it not work for me
> 
> Perhaps another plugin?

Seems that it is, in fact, being caused by another plugin. Can't reproduce the issue with the minimal init

famiu:
> > huh... then I wonder what's making it not work for me
> 
> Perhaps another plugin?

Wait, no, it's not being caused by another plugin. The issue only occurs if I `cd` into a git repo and start nvim from there

PR #89, opened 2021-09-19
feat: allow all component values to take winid

### LINKED ISSUE #84: [Feature]Add winid to hl function
Hi, thanks for the plugin. I am currently trying to give different highlight to the filename depending if the file is modified. However, unlike provider which takes in `winid` as an argument, it seems like highlight function changes the highlighting globally for all statusbars for the buffers.

I wish the highlight function can take in a winid as well, so that I am only modifying the statusbar for one particular buffer.

So inorder to achieve the different coloring, I use the `vim.bo.modified`. Here is my code.

```lua
table.insert(components.active, {})
table.insert(components.active, {})

table.insert(components.inactive, {})
table.insert(components.inactive, {})

local split = function(inputstr, sep)
	if sep == nil then
		sep = "%s"
	end
	local t = {}
	for str in string.gmatch(inputstr, "([^" .. sep .. "]+)") do
		table.insert(t, str)
	end
	return t
end

local ins_left = function(component)
	table.insert(components.active[1], component)
end
local ins_right = function(component)
	table.insert(components.active[#components.active], component)
end

local short_ins_left = function(component)
	table.insert(components.inactive[1], component)
end
local short_ins_right = function(component)
	table.insert(components.inactive[#components.active], component)
end
local file_path = {
	provider = function(_, winid)
		local filepath = vim.fn.fnamemodify(vim.api.nvim_buf_get_name(vim.api.nvim_win_get_buf(winid)), ":~:.")
		local splitpath = split(filepath, "/")
		splitpath[#splitpath] = ""
		filepath = table.concat(splitpath, "/")
		return filepath
	end,
	hl = function()
		local res = {
			fg = colors.yellow,
			style = "bold",
		}
		if vim.bo.modifiable and vim.bo.modified then
			res.fg = colors.red
		end
		return res
	end,
}
ins_left(file_path)

local file_name = {
	provider = function(_, winid)
		return fn.fnamemodify(vim.api.nvim_buf_get_name(vim.api.nvim_win_get_buf(winid)), ":t")
	end,
	hl = function()
		local res = {
			fg = colors.fg,
			style = "bold",
		}
		if vim.bo.modifiable and vim.bo.modified then
			res.fg = colors.red
		end
		return res
	end,
	icon = "",
}
ins_left(file_name)

short_ins_left(file_path)
short_ins_left(file_name)
```
However, this code not only modify the current buffer statusline, it also modifies other buffers as well. I think this can be solved with `vim.bo[bufnr].modified`, hence I wish there is an option for `winid`.

**Additional context**
Before modifying feline.lua buffer.
image
After modifying feline.lua buffer.
image
After jumping to galaxyline buffer and jumping back to feline.lua
image

So the inactive buffer gets updated when trigger events happen, and the inactive buffer will be red when the active buffer is modified.

famiu:
Closes #84 

Since this adds a lot to the table, I think it needs to be tested well before merging, but from the testing I've done, it seems to work fine

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,7 +260,7 @@ components.active[3][2] = {
 
 **NOTE:** If you use the index instead of table.insert, remember to put the correct index. Also keep in mind that unlike most other programming languages, Lua indices start at `1` instead of `0`.
 
-You can customize each component to your liking. Most values that a component requires can also use a function. However, these functions cannot take any arguments, with the exception of the `provider` and `enabled` values, which can take arguments (more about that below).
+You can customize each component to your liking. Most values that a component requires can also use a function. These functions can take either no arguments or exactly one argument, the window id (`winid`) of the window for which the statusline is being generated. However, the `provider` value is an exception because it can take two arguments (more on that below).
 
 Feline will automatically evaluate the function if it is one. But in case a function is provided, the type of the value the function returns must be the same as the type of value required by the component. For example, since `enabled` requires a boolean value, if you set it to a function, the function must also return a boolean value.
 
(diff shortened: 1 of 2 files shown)


ram02z:
As long as this satisfies the feature request, I think merging is fine since it is to the develop branch.

famiu:
> As long as this satisfies the feature request, I think merging is fine since it is to the develop branch.

That's true, I suppose

PR #94, opened 2021-09-19
feat: add support for named components

### LINKED ISSUE #75: [Feature] Named components?
I was thinking of adding named components to Feline's. So basically, components could have an extra `name` value, which could be used to access them directly from the components table without having to remember their index. It'd also help prevent code duplication in the plugin since `noicons` preset is basically the `default` preset with a few changes, we'd be able to just directly use the `default` preset and make a few modifications to it using the component names to prevent code duplication. It'd also allow people to easily modify components without having to remember the index.

The problem is, I haven't found a single way to implement them that isn't horribly ineffecient (eg: something that doesn't involve searching every component whenever a name is used for access). I tried caching the names and using metatables for access but even then it isn't foolproof since `newindex` only works for values that already don't exist, and using a proxy table prevents the table from being iterated using `ipairs`, which is something that the statusline generator needs to do. Thoughts on it would be appreciated.

famiu:
Closes #75 

Still need to add docs. Implementation is a little bit hacky since it makes extensive use of metatables, but it seems to work fine.

Would like if it was tested by others as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,8 @@ Feline will automatically evaluate the function if it is one. But in case a func
 
 Note that you can omit all of the component values except `provider`, in which case the defaults would be used instead. A component can have the following values:
 
+- `name` (string): If provided, it allows easily accessing the component for modifying it without having to know or remember its indices using `require('feline').components[component_name]`. If that's not needed, however, then the component name can be omitted without any issues.
+
 - `provider` (string or function): If it's a string, it represents the text to show. If it's a function, it must return a string when called. As a function it may also optionally return an `icon` component alongside the string when called, which would represent the provider's icon, possibly along with the icon highlight group configuration. The function can take either no arguments, or one argument which would contain the component itself, or it can take two arguments, the component and the window handler of the window for which the statusline is being generated.
 
 ```lua
--- a/README.md
+++ b/README.md
@@ -266,7 +266,7 @@ Feline will automatically evaluate the function if it is one. But in case a func
 
 Note that you can omit all of the component values except `provider`, in which case the defaults would be used instead. A component can have the following values:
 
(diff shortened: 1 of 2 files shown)


famiu:
I wonder if it's a feature really worth adding

README.md
```diff
@@ -266,6 +266,8 @@ Feline will automatically evaluate the function if it is one. But in case a func
 
 Note that you can omit all of the component values except `provider`, in which case the defaults would be used instead. A component can have the following values:
 
+- `name` (string): If provided, it allows easily accessing the component for modifying it without having to know or remember its indices using `require('feline').components[component_name]`. If that's not needed, however, then the component name can be omitted without any issues.
```
ram02z, line 269:
```suggestion
- `name` (string): If provided, it allows the component to be easily modified without having to know or remember its indice using `require('feline').components[component_name]`. If that's not needed, then the component name can be omitted without any issues.
```

ram02z:
This could be a cool addition, however, I don't think I would use this feature since my statusline doesn't really have duplicate features. Still, I think due to the fact that feline has two configured statuslines, active and inactive, this could help with deduplication with more advanced configurations.

famiu:
> This could be a cool addition, however, I don't think I would use this feature since my statusline doesn't really have duplicate features (can't test). Still, I think due to the fact that feline has two configured statuslines, active and inactive, this could help with deduplication with more advanced configurations.

I was actually thinking of adding an `inherit` value for components so they can inherit a component with the name, though I suppose the same can be done manually too so idk if that'd be a good idea

famiu:
Honestly, the more I try to perfect the implementation, the more complex it gets, and honestly, what it adds is not worth the complexity imo. So I'm going to have to abort this

PR #98, opened 2021-09-21
Named components (take 2)

famiu:
I decided to give named components another shot and actually managed to perfect it this time. The question this time, though, is whether the features it adds (possibility for better error messages, easy access to components) worth the complexity it adds to the codebase. @ram02z thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,7 +266,7 @@ Feline will automatically evaluate the function if it is one. But in case a func
 
 Note that you can omit all of the component values except `provider`, in which case the defaults would be used instead. A component can have the following values:
 
-- `name` (string): If provided, it allows the component to be easily modified without having to know or remember its index using `require('feline').components[component_name]`. In case of an error while parsing a component, it also allows Feline to point out which component the error is coming from if the component has a name. If you don't need any of that, then the component name can be omitted without any issues.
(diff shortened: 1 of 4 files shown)


famiu:
After doing some benchmark, it also seems this has a rather heavy performance cost

ram02z:
Like I said in the first PR, I probably won't be using this feature. However, I did give this PR a try and I can't seem to get it to work. I gave one of my providers a name, `vi`, and then tried to add it the inactive statusline like so, `table.insert(components.inactive[1], require('feline').components['vi'])`. Not sure if that is the intended usecase, but, I get an error that `components` is a nil value.

famiu:
> Like I said in the first PR, I probably won't be using this feature. However, I did give this PR a try and I can't seem to get it to work. I gave one of my providers a name, `vi`, and then tried to add it the inactive statusline like so, `table.insert(components.inactive[1], require('feline').components['vi'])`. Not sure if that is the intended usecase, but, I get an error that `components` is a nil value.

You actually need to use Lua indices instead of table.insert, also you need to use `require('feline').components['vi'].copy()` and you also need to change the name since two components can't have the same name. But yes, there's a lot of gotchas associated with it, so honestly I'm going to have to abandon this anyway, despite having perfected auto-updating the cache after a name change.

Or, we could remove the whole access-component-by-name thing but still keep component names in order to show better error messages.

Issue #96, opened 2021-09-20
[Feature] Smart component truncation


famiu:
I'm willing to give this another shot. But first, we'd have to plan how truncation actually works. How do we decide when to truncate, which components to truncate, and where to truncate the components, and how do we allow the provider to affect the truncation? We need to plan out a good blueprint for truncation before we can implement it, so I'd like the input of as many people as possible.

@ram02z you seemed to have ideas for truncation as well, I'd be interested in hearing them out again. @Melkster how would you want truncation to work?

ram02z:
The way I would approach truncation would require a slight refactor and breaking changes for the provider key. In order for the truncation to be useful, it will have to be a core feature, so you can't really opt out. Firstly, we would have to store the state of the statusline. The main elements we would require for truncation would be the remaining length of the section (assuming that sections are equal in length), the current window size and the truncation priority. The priority system will be important as it will allow us to know what components need to be truncated, therefore, it should have its own key for components where the user can change the priority of component. If the component, doesn't have a priority set, it will just be left to right (index 1..N). Now, in order for the truncation to be useful, I propose three states: normal, short and hidden. This would require a seperate provider for the short version, if the user omits the short provider, the provider will just get hidden. Therefore, feline will first compare the current window length and the remaining section length and use that to either show the full provider, the short provider or just hide it entirely. I think this method is better than the max_length method because that is quite naive since it just splits the string if it is greater than N characters. Also, I don't think you have to worry about the fact that neovim doesn't have a `WinResized` autocommand since the statusline is smart enough to update when the window is resized (you can try this out by using the `enabled` key to query the window size)

famiu:
[...]

PR #102, opened 2021-09-23
refactor: separate provider options from component values

famiu:
BREAKING CHANGE: The arguments for provider functions have been changed.

Prerequisite for component truncation (#96). This seems to be working fine. Would still like a review from @ram02z


famiu:
Since this seems to work fine, I'll merge this now so I can start work on truncation

Issue #106, opened 2021-09-25
[Bug] file_info doesn't have icon when setting `opts`


FollieHiyuki:
**Neovim version**
```
NVIM v0.5.0
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/lib/ccache/bin/cc -fstack-clash-protection -D_FORTIFY_SOURCE=2 -mtune=generic -O2  -g -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/builddir/neovim-0.5.0/build/config -I/builddir/neovim-0.5.0/src -I/usr/include -I/builddir/neovim-0.5.0/build/src/nvim/auto -I/builddir/neovim-0.5.0/build/include
Compiled by void-buildslave@a-hel-fi

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```

**Feline version**
Up-to-date `master` branch

**Describe the bug**
When setting `opts` for `file_info` provider, the file icon doesn't appear, even though `icon` is not set to anything.

**To Reproduce**
Setting `file_info` with `opts`. For example:
```lua
    require('feline').setup {
        components = {
            active = {
                {
                    {
                        provider = {
                            name = 'file_info',
                            opts = {
                                -- Set anything here
                                type = 'unique'
                            }
                        },
                        -- Don't set icon
                        hl = { fg = '#61afef', style = 'bold'},
                        right_sep = ' '
                    }
                },
                {}, {}
            },
            inactive = {}
        }
    }
```

**Expected behavior**
There is a file icon appear before the file name (the same as normally setting `provider = 'file_info'`)

**Does this error occur in the minimal init file?**
Yes

**Provide modified minimal_init.lua**
See the reproduce part above

**Screenshots**

[...]

PR #108, opened 2021-09-25
Fix feline ignoring component.opts

xeluxee:
`opts` in component was ignored, so `file_info` options were ineffective.


famiu:
This is wrong because `component` doesn't have any `opts` value. Closing this

xeluxee:
Sorry I misunderstood it due to things changing in recent updates
Initially I made this PR not to fix #106, but to fix `opts` settings being ignored.
Then I read USAGE.md and I discovered that `provider` has to be configured in a different way :sweat_smile:

PR #110, opened 2021-09-26
feat: add smart component truncation

### LINKED ISSUE #96: [Feature] Smart component truncation

famiu:
Closes #96 

Seems to be working fine. Though I'd like @ram02z to review the code (and try to make it simpler if possible). I'd also like it to be well-tested before merge.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -169,6 +169,62 @@ end
 
 If you omit the provider value, it will be set to an empty string. A component with no provider or an empty provider may be useful for things like [applying a highlight to section gaps](#highlight-section-gaps) or just having an icon or separator as a component.
 
+#### Component short provider
+
+Feline has an automatic smart truncation system where components can be automatically truncated if the statusline doesn't fit within the window. In order to make use of this truncation system, you have to define the `short_provider` component value.
+
+`short_provider` works just like the `provider` value, but is activated only when the component is being truncated due to the statusline not fitting within the window. `short_provider` is independent from the `provider` value so it can be a different provider altogether, or it can be a shortened version of the same provider or the same provider but with a different `opts` value. For example:
+
+```lua
+-- In this component, short provider uses same provider but with different opts
+local file_info_component = {
+    provider = {
+        name = 'file_info',
+        opts = {
+            type = 'full-path'
+        }
+    },
+    short_provider = {
+        name = 'file_info',
+        opts = {
+            type = 'short-path'
+        }
+    }
+}
+
+-- Short provider can also be an independent value / function
+local my_component = {
+    provider = 'loooooooooooooooong',
+    short_provider = 'short'
+}
+```
+
+Feline doesn't set `short_provider` to any component by default, so it must be provided manually.
+
+#### Hide components during truncation
+
+If you wish to allow Feline to hide a component entirely if necessary during truncation, you may set the `truncate_hide` component value to `true`. By default, `truncate_hide` is `false` for every component.
+
+#### Component priority
+
+When components are being truncated by Feline, you can choose to give some components a higher priority over the other components. The `priority` component value just takes a number. By default, the priority of a component is `0`. Components are truncated in ascending order of priority. So components with lower priority are truncated first, while components with higher priority are truncated later on. For example:
+
+```lua
+-- This component has the default priority
+local my_component = {
+    provider = 'loooooooooooooooong',
+    short_provider = 'short'
+}
+-- This component has a higher priority, so it will be truncated after the previous component
+local high_priority_component = {
+    provider = 'long provider with high priority',
+    short_provider = 'short',
+    priority = 1
+}
+```
+
+Priority can also be set to a negative number, which can be used to make a component be truncated earlier than the ones with default priority.
+
 #### Conditionally enable components
 
 The `enabled` value of a component can be a boolean or function. This value determines if the component is enabled or not. If false, the component is not shown in the statusline. If it's a function, it can take either the window handler as an argument, or it can take no arguments. For example:
--- a/lua/feline/generator.lua
+++ b/lua/feline/generator.lua
@@ -333,6 +333,108 @@ local function parse_component(component, winid, use_short_provider)
     }
 end
 
+-- Parse statusline sections and truncate the components when necessary
+local function parse_statusline_sections(sections, statusline_type, winid)
+    -- Parse all components in the sections and also store the indices of every component so
+    -- both the sections and parsed_sections tables can be accessed later on after components
+    -- are sorted in order of priority
+    -- Also calculate statusline length while doing all of that
+    local parsed_sections = {}
+    local component_indices = {}
+    local statusline_length = 0
+
+    for i, section in ipairs(sections) do
+        parsed_sections[i] = {}
+
+        for j, component in ipairs(section) do
+            local ok, result = pcall(parse_component, component, winid)
+
+            if not ok then
+                api.nvim_err_writeln(string.format(
+                    "Feline: error while processing component number %d on section %d "..
+                    "of type '%s': %s",
+                    j, i, statusline_type, result
+                ))
+
+                result = { str = '', len = 0 }
+            end
+
+            parsed_sections[i][j] = result
+            component_indices[#component_indices+1] = {i, j}
+            statusline_length = statusline_length + result.len
+        end
+    end
+
+    local win_width = api.nvim_win_get_width(winid)
+
+    -- If statusline length is larger than the window width, sort the component indices
+    -- in ascending order of priority of the components they refer to
+    -- Then truncate the components one by one using by their short_provider or hiding them
+    -- entirely until the statusline fits within the window
+    if statusline_length > win_width then
+        table.sort(component_indices, function(a, b)
+            -- Access the original component through the sections table using the indices
+            local first_component_priority = sections[a[1]][a[2]].priority or 0
+            local second_component_priority = sections[b[1]][b[2]].priority or 0
+
+            return first_component_priority < second_component_priority
+        end)
+
+        for _, indices in ipairs(component_indices) do
+            -- Access the original and parsed values using the indices
+            local component = sections[indices[1]][indices[2]]
+            local parsed_component = parsed_sections[indices[1]][indices[2]]
+
+            -- If short_provider exists, use it
+            if component.short_provider then
+                -- Get new parsed component value using the short_provider and calculate the
+                -- length difference between the two values, and if it's greater than 0, use
+                -- the new value instead of the old one
+                local parsed_component_new = parse_component(component, winid, true)
+                local length_difference = parsed_component.len - parsed_component_new.len
+
+                if length_difference > 0 then
+                    -- Update statusline length and replace old parsed value with new one
+                    statusline_length = statusline_length - length_difference
+                    parsed_sections[indices[1]][indices[2]] = parsed_component_new
+                end
+            end
+
+            if statusline_length <= win_width then break end
+        end
+
+        -- If statusline still doesn't fit, start removing components with truncate_hide
+        if statusline_length > win_width then
+            for _, indices in ipairs(component_indices) do
+                if sections[indices[1]][indices[2]].truncate_hide then
+                    statusline_length = statusline_length -
+                        parsed_sections[indices[1]][indices[2]].len
+
+                    parsed_sections[indices[1]][indices[2]] = {str = '', len = 0}
+                end
+
+                if statusline_length <= win_width then break end
+            end
+        end
+    end
+
+    -- Concatenate all components in each section to get a string for each section
+    local section_strs = {}
+
+    for i, parsed_section in ipairs(parsed_sections) do
+        local component_strs = {}
+
+        for j, parsed_component in ipairs(parsed_section) do
+            component_strs[j] = parsed_component.str
+        end
+
+        section_strs[i] = table.concat(component_strs)
+    end
+
+    -- Finally, concatenate all sections to get the statusline string
+    return table.concat(section_strs, '%=')
+end
+
 -- Generate statusline by parsing all components and return a string
 function M.generate_statusline(winid)
     -- Generate default highlights for the statusline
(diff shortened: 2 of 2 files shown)


famiu:
After doing some benchmarks I can see that it makes the statusline generation time around 15% higher than before

famiu:
I wonder what happens when a window is resized causing a component in an inactive window to require truncation or allowing a component in an inactive window to not be truncated. There's no `WinResized` autocommand so it's impossible to update inactive statuslines in those cases. The next best thing is `VimResized` but the windows can still be resized without triggering `VimResized`

ram02z:
> 
> I wonder what happens when a window is resized causing a component in an inactive window to require truncation or allowing a component in an inactive window to not be truncated. There's no `WinResized` autocommand so it's impossible to update inactive statuslines in those cases. The next best thing is `VimResized` but the windows can still be resized without triggering `VimResized`

You are right, this is a limitation of neovim since there is no `WinResized` autocommand and the only linked [PR ](https://github.com/neovim/neovim/pull/13589) has been stale.

I quickly tried the file_info example you added to the README and the switching between the normal provider and short provider works well, but it doesn't seem to hide the short provider if it is still too big (took a quick look at the changes and I don't think it was implemented). I don't think it's that big of a deal though, since the user could use the enabled key for that.

I currently don't have much time to thoroughly test and review this properly, however, I will do as soon as I can.

famiu:
> I quickly tried the file_info example you added to the README and the switching between the normal provider and short provider works well, but it doesn't seem to hide the short provider if it is still too big (took a quick look at the changes and I don't think it was implemented). I don't think it's that big of a deal though, since the user could use the enabled key for that.

Question, should it start removing components only after all components with `short_provider` are using their `short_provider` regardless of whether the component with `short_provider` has higher priority, or should it remove the component if it doesn't have a `short_provider` when truncating components in order of priority?

ram02z:
> 
> Question, should it start removing components only after all components with `short_provider` are using their `short_provider` regardless of whether the component with `short_provider` has higher priority, or should it remove the component if it doesn't have a `short_provider` when truncating components in order of priority?

Not necessarily, we should only care about the length of providers and the priority. Same reason why I think it is up to the user to ensure that the short provider is actually shorter than the normal provider. I think that components that don't have a short provider should be hidden when truncated in order of priority and when they do have a short provider, feline should first attempt to fallback to the short provider, but if the short provider's length still surpasses the section limit, it should be hidden. Perhaps, hiding the components should be an opt-in feature though by introducing a key to feline's setup or maybe to each component; whatever you think is better. I am suggesting that it should be opt-in because some users may not want their component's hidden unless they explicitly define that behaviour themselves.

famiu:
> > Question, should it start removing components only after all components with `short_provider` are using their `short_provider` regardless of whether the component with `short_provider` has higher priority, or should it remove the component if it doesn't have a `short_provider` when truncating components in order of priority?
> 
> Not necessarily, we should only care about the length and priority of the providers. Same reason why I think it is up to the user to ensure that the short provider is actually shorter than the normal provider. I think that components that don't have a short provider should be hidden when truncated in order of priority and when they do have a short provider, feline should first attempt to fallback to the short provider, but if the short provider's length still surpasses the section limit, it should be hidden. Perhaps, hiding the components should be an opt-in feature though by introducing a key to feline's setup or maybe to each component; whatever you think is better. I am suggesting that it should be opt-in because some users may not want their component's hidden unless they explicitly define that behaviour themselves.

I'm going to add an `no_hide` option then

famiu:
> > > Question, should it start removing components only after all components with `short_provider` are using their `short_provider` regardless of whether the component with `short_provider` has higher priority, or should it remove the component if it doesn't have a `short_provider` when truncating components in order of priority?
> > 
> > 
> > Not necessarily, we should only care about the length and priority of the providers. Same reason why I think it is up to the user to ensure that the short provider is actually shorter than the normal provider. I think that components that don't have a short provider should be hidden when truncated in order of priority and when they do have a short provider, feline should first attempt to fallback to the short provider, but if the short provider's length still surpasses the section limit, it should be hidden. Perhaps, hiding the components should be an opt-in feature though by introducing a key to feline's setup or maybe to each component; whatever you think is better. I am suggesting that it should be opt-in because some users may not want their component's hidden unless they explicitly define that behaviour themselves.
> 
> I'm going to add an `no_hide` option then

Actually, thinking more about it, I think it'd just be better if components aren't hidden at all, best to leave that to the user

ram02z:
> 
> Actually, thinking more about it, I think it'd just be better if components aren't hidden at all, best to leave that to the user

I agree with that as well, which I suggested it should be opt-in behaviour (disabled by default). It might as well get implemented since it will make use of feline's priority system, so it would be redundant if the user tried to replicate that behaviour in their own config.

famiu:
> > Actually, thinking more about it, I think it'd just be better if components aren't hidden at all, best to leave that to the user
> 
> I agree with that as well, which I suggested it should be opt-in behaviour (disabled by default). It might as well get implemented since it will make use of feline's priority system, so it would be redundant if the user tried to replicate that behaviour in their own config.

In theory, setting `short_provider = ''` should have the same effect, though separators and icons with `always_visible` can still appear in that case

ram02z:
> 
> But if we do make it an opt-in feature either way, I wonder if it'd be better to have one setting for the entire statusline through `setup` or have it individually for each component.

If the implementation isn't complicated, per component customisation is preferred imo.

famiu:
So I finally added the feature to hide components entirely. Again, this seems to be working fine.

But even though it's a cool feature and all, I can't say I'm a fan of the performance degrade this feature brings, not to mention the increased complexity of the code.

famiu:
> So I finally added the feature to hide components entirely. Again, this seems to be working fine.
> 
> But even though it's a cool feature and all, I can't say I'm a fan of the performance degrade this feature brings, not to mention the increased complexity of the code.

After doing some benchmarks again though it seems that the performance downgrade is not as bad as I had first seen, which is relieving. The code could still use some simplifying though, except I don't really have much of an idea on how to approach that

famiu:
Simplified the code a little bit, not sure how readable it is now though. @ram02z when you have time I'd like to know if you have trouble understanding the code

lua/feline/generator.lua
```diff
@@ -143,106 +145,133 @@ local function get_hlname(hl, parent_hl)
     return hlname
 end
 
--- Parse component seperator
+-- Parse component seperator to return parsed string and length
 -- By default, foreground color of separator is background color of parent
 -- and background color is set to default background color
 local function parse_sep(sep, parent_bg, is_component_empty)
-    if sep == nil then return '' end
+    if sep == nil then  return {str = '', len = 0} end
```
ram02z, line 152: `:s/then  return/then return/g`

lua/feline/generator.lua
```diff
@@ -328,17 +347,109 @@ function M.generate_statusline(winid)
...
+        local win_width = api.nvim_win_get_width(winid)
+
+        -- If statusline length is larger than the window width, sort the component indices
+        -- in ascending order of priority of the components they refer to
+        -- Then truncate the components one by one using by their short_provider or hiding them
+        -- entirely until the statusline fits within the window
+        if statusline_length > win_width then
+            table.sort(component_indices, function(a, b)
+                -- Access the original component through the sections table using the indices
+                return (sections[a[1]][a[2]].priority or 0) < (sections[b[1]][b[2]].priority or 0)
```
ram02z, line 391: This line is abit too complicated. I would maybe create two variables and compared them to make it more readable.

lua/feline/generator.lua
```diff
@@ -328,17 +347,109 @@ function M.generate_statusline(winid)
...
+                    local length_difference = parsed_component.len - parsed_component_new.len
+
+                    if length_difference > 0 then
+                        -- Update statusline length and replace old parsed value with new one
+                        statusline_length = statusline_length - length_difference
+                        parsed_sections[indices[1]][indices[2]] = parsed_component_new
+                    end
+                elseif component.truncate_hide then
+                    statusline_length = statusline_length - parsed_component.len
+                    parsed_sections[indices[1]][indices[2]] = {str = '', len = 0}
```
ram02z, line 415:
```suggestion

```

This isn't needed since the check happens after anyways. Also causes an issue when you have a component with no short provider, high priority and truncate_hide set to true. It will still get hidden even though it has a priority higher than a component with a short provider.

USAGE.md
```diff
@@ -167,6 +167,66 @@ provider = function(_, _, opts)
...
+-- Short provider can also be an independent value / function
+local my_component = {
+    provider = 'loooooooooooooooong',
+    short_provider = 'short'
+}
+```
+
+Feline doesn't set `short_provider` to any component by default, so it must be provided manually.
+
+#### Hiding components during truncation
```
ram02z, line 202: This will get generated as `feline-Hiding-components-during-truncation`, so maybe simplify the heading to `Hide truncated component`

ram02z (changes requested):
LGTM. Once neovim releases a `WinResized` autocommand most of the truncation logic in `generate_statusline` can be to a function triggered by it. I think you should definitely CC the feline users that expressed interested in the truncation to also test this branch, but from what I have tested so far, it seems to working as intended.

ram02z:
> 
> 
> Simplified the code a little bit, not sure how readable it is now though. @ram02z when you have time I'd like to know if you have trouble understanding the code

The code itself is not difficult to understand and the comments are useful too. The only issue I think may that debugging truncation issues may be difficult due to how tables like `parsed_sections` can get abit convoluted.

famiu:
> LGTM. Once neovim releases a `WinResized` autocommand most of the truncation logic in `generate_statusline` can be moved to a function triggered by it.

Actually, I don't think so, since it needs to modify the generated statusline string directly.

famiu:
> The code itself is not difficult to understand and the comments are useful too. The only issue I think may that debugging truncation issues may be difficult due to how tables like `parsed_sections` can get abit convoluted.

Any suggestions to fix that?

lua/feline/generator.lua
famiu: Is it better now?

lua/feline/generator.lua
famiu: That isn't supposed to happen because components are always hidden in order of priority. Also the check that happens afterwards hides all components with `truncate_hide` set to true, even if they have a short provider, which should only happen if all other components are using their `short_provider`

lua/feline/generator.lua
ram02z: Yeah, better now.

lua/feline/generator.lua
ram02z: Well, for one there is no point of checking the `truncate_hide` value for components without a short_provider because they would have been hidden by the code I suggested to removed anyway. Anyway, you should be able to reproduce the bug with this [config](https://gist.github.com/ram02z/34d6f2fae2c1a35b455ea480ea2da726). Just keep making the active smaller and you will see that the higher priority component is still hidden first.

famiu: > Well, for one there is no point of checking the `truncate_hide` value for components without a short_provider because they would have been hidden by the code I suggested to be removed anyway. Anyway, you should be able to reproduce the bug with this [config](https://gist.github.com/ram02z/34d6f2fae2c1a35b455ea480ea2da726). Just keep making the active smaller and you will see that the higher priority component is still hidden first.

Ah, I see what you mean now. That behavior was intended actually, but thinking more about it, it makes more sense to remove it

ram02z:
> 
> 
> > LGTM. Once neovim releases a `WinResized` autocommand most of the truncation logic in `generate_statusline` can be moved to a function triggered by it.
> 
> Actually, I don't think so, since it needs to modify the generated statusline string directly.

True, but that autocommand at the bare minmum will allow for truncation to be visible for inactive windows.

> 
> 
> > The code itself is not difficult to understand and the comments are useful too. The only issue I think may that debugging truncation issues may be difficult due to how tables like `parsed_sections` can get abit convoluted.
> 
> Any suggestions to fix that?

Honestly, not sure because the `sections` table in itself needs two indices to access a specific component. I don't even think named components would help.

famiu:
> > > LGTM. Once neovim releases a `WinResized` autocommand most of the truncation logic in `generate_statusline` can be moved to a function triggered by it.
> > 
> > 
> > Actually, I don't think so, since it needs to modify the generated statusline string directly.
> 
> True, but that autocommand at the bare minmum will allow for truncation to be visible for inactive windows.
> 
> > > The code itself is not difficult to understand and the comments are useful too. The only issue I think may that debugging truncation issues may be difficult due to how tables like `parsed_sections` can get abit convoluted.
> > 
> > 
> > Any suggestions to fix that?
> 
> Honestly, not sure because the `sections` table in itself needs two indices to access a specific component. I don't even think named components would help.

Yeah, that's an issue. If only Lua allowed taking variable references...

famiu:
This doesn't seem to work anymore after I rebased it against the develop branch, not sure why

famiu:
This seems to be caused by #120

famiu:
Removing the `vim.schedule` wrapper does seem to fix both #119 and also allow truncation, but then it causes the NvimTree problem...

famiu:
> Removing the `vim.schedule` wrapper does seem to fix both #119 and also allow truncation to work, but then it causes the NvimTree problem...

Honestly, at this point I'd suggest removing the `vim.schedule()` part and ask @xeluxee to open an issue on NvimTree and ask NvimTree to add an autocmd for when it's done doing its work, so @xeluxee can then add that to their update triggers. As I see it now, `vim.schedule()` is kind of unpredictable and a very hacky "fix", so maybe it's better to remove it altogether.

ram02z:
> 
> Honestly, at this point I'd suggest removing the `vim.schedule()` part and ask @xeluxee to open an issue on NvimTree and ask NvimTree to add an autocmd for when it's done doing its work, so @xeluxee can then add that to their update triggers. As I see it now, `vim.schedule()` is kind of unpredictable and a very hacky "fix", so maybe it's better to remove it altogether.

I don't personally have a problem when `vim.schedule` is not used to update the windows, and `NerdTree` is the only plugin that we noticed misbehaved.

famiu:
> I am wondering if #120 is still required once `vim.schedule` is removed, feline fallbacks to the global statusline anyway.

It's not

famiu:
Since this seems to be working fine I'm going to merge this now. Make a new issue if this causes any problems

famiu:
@ram02z it seems that truncation doesn't work correctly still because in Vim's statusline, in order to have a literal `%` character in it, you have to escape it with another `%`, which is what the `line_percentage` provider uses. As a result it gives an incorrect length for it. Not sure how to fix it

famiu:
I might have to remove truncation. It still seems to have a lot of flaws (such as the one mentioned above), and it's not really very reliable, plus it seems to have a massive performance hit as well and it's kind of impossible to make it zero-cost.

famiu:
I've opened an issue in the Neovim repo that'd make the process much easier: https://github.com/neovim/neovim/issues/15849

But until that's implemented, I'm strongly in favor of removing truncation. No truncation is better than half-assed truncation imo.

famiu:
I'm going to remove truncation for now until either the feature gets implemented in Neovim or until I can find a more performant and reliable way to get the statusline length

shadmansaleh:
You can use `build_stl_str_hl` function through lua ffi to get expanded result.

Even if neovim implements expanding stl  expressions it'll still be really costly . As it can't really be lighter then what neovim currently does while drawing statusline.

famiu:
> You can use `build_stl_str_hl` function through lua ffi to get expanded result.

Oh? I'm not sure how to do that though

> Even if neovim implements expanding stl expressions it'll still be costly . As it can't really be lighter then what neovim currently does while drawing statusline.

But at least then it'll be more accurate

shadmansaleh:
It'll be something like
```lua
local ffi = require('ffi')

local M = {}
ffi.cdef([[
typedef unsigned char char_u;
typedef struct window_S win_T;
extern win_T   *curwin;
typedef struct {} stl_hlrec_t;
typedef struct {} StlClickDefinition;
typedef struct {} StlClickRecord;
int build_stl_str_hl(
    win_T *wp,
    char_u *out,
    size_t outlen,
    char_u *fmt,
    int use_sandbox,
    char_u fillchar,
    int maxwidth,
    stl_hlrec_t **hltab,
    StlClickRecord **tabtab
);
]])

function M.expand_stl_expr(stl_fmt)
    local stlbuf = ffi.new('char_u [?]', 256)
    local fmt = ffi.cast('char_u *', stl_fmt)
    ffi.C.build_stl_str_hl(
        ffi.C.curwin,
        stlbuf,
        256,
        fmt,
        0,
        ffi.cast('char_u', 0x20),
        vim.fn.winwidth(0),
        nil,
        nil
    )
    return ffi.string(stlbuf)
end

return M
```

Note: Untested code .

famiu:
> > Oh? I'm not sure how to do that though
> 
> It'll be something like this
> 
> ```lua
> local ffi = require('ffi')
> 
> local M = {}
> 
> ffi.cdef [[
> typedef unsigned char char_u;
> typedef struct window_S win_T;
> extern win_T   *curwin;
> typedef struct {} stl_hlrec_t;
> typedef struct {} StlClickDefinition;
> typedef struct {} StlClickRecord;
> int build_stl_str_hl(
>     win_T *wp,
>     char_u *out,
>     size_t outlen,
>     char_u *fmt,
>     int use_sandbox,
>     char_u fillchar,
>     int maxwidth,
>     stl_hlrec_t **hltab,
>     StlClickRecord **tabtab
> );
> ]]
> 
> function M.expand_stl_expr(stl_fmt)
>     local stlbuf = ffi.new('char_u [?]', 256)
>     local fmt = ffi.cast('char_u *', stl_fmt)
>     ffi.C.build_stl_str_hl(
>         ffi.C.curwin,
>         stlbuf,
>         256,
>         fmt,
>         0,
>         ffi.cast('char_u', 0x20),
>         vim.fn.winwidth(0),
>         nil,
>         nil
>     )
>     return ffi.string(stlbuf)
> end
> 
> return M
> ```
> 
> Note: Untested code .

Oh wow, this actually seems to work pretty well. Thanks a lot for this!

PR #115, opened 2021-09-26
Allow separator hl keys to be functions

rktjmp:
Might be by design, or omission. Currently it's inconsistent with the component hl key.


famiu:
I don't think it needs to be a function personally since separator itself can be a function, but if there's a good argument for it then I might be interested in adding it

rktjmp:
Consistency would be the only argument, but I agree.

famiu:
By the way, both the `str` and `hl` keys of both separators and icons can be functions now

rktjmp:
Feline might be the most *functional* status line around :grimacing:

PR #123, opened 2021-09-29
feat: allow truncation to work with Vim statusline modifiers

famiu:
Previously, truncation couldn't work with Vim statusline modifiers due to the length of the component string not being calculated properly if it had a Vim statusline modifier. This commit introduces a `get_statusline_str_length()` function which manually parses the component strings to get their proper length.

The statusline should work fine but `get_statusline_str_length()` only supports highlight modifiers so far, so the function needs more work done to be finished.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -229,8 +229,6 @@ local high_priority_component = {
 
 Priority can also be set to a negative number, which can be used to make a component be truncated earlier than the ones with default priority.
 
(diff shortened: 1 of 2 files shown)


famiu:
It just occured to me that Vim also allows function calls and such as statusline modifiers, so this will probably never work. But I still think the approach taken here is a lot better than returning the length with every function

PR #132, opened 2021-10-02
refactor: re-add smart component truncation

famiu:
Now fully supports VIm statusline expressions. Thanks to @shadmansaleh for assisting with the FFI code required for that. This should work perfectly fine. The only issue is that it causes a 53% performance downgrade, which is a pretty big issue. The downgrade is not really noticable to me but not sure about others. Not a fan of it using FFI to access Neovim's internals either, mostly because I don't have much knowledge about Neovim internals so it'd be a pain to troubleshoot a problem if the source of the problem is the FFI code.

### DIFF
--- a/lua/feline/statusline_ffi.lua
+++ b/lua/feline/statusline_ffi.lua
@@ -1,0 +1,72 @@
+-- This module provides an interface to Neovim's statusline generator using Lua FFI
+local M = {}
+
+local ffi = require('ffi')
+
+-- Definitions required to use Neovim's build_stl_str_hl function to expand statusline expressions
+-- And also other definitions used by this module
+ffi.cdef [[
+typedef unsigned char char_u;
+typedef struct window_S win_T;
+typedef struct {} stl_hlrec_t;
+typedef struct {} StlClickRecord;
+
+extern win_T *curwin;
+
+int build_stl_str_hl(
+    win_T *wp,
+    char_u *out,
+    size_t outlen,
+    char_u *fmt,
+    int use_sandbox,
+    char_u fillchar,
+    int maxwidth,
+    stl_hlrec_t **hltab,
+    StlClickRecord **tabtab
+);
+]]
+
+-- Used CType values stored in a local variable to avoid redefining them and improve performance
+local char_u_buf_t = ffi.typeof('char_u[?]')
+local char_u_str_t = ffi.typeof('char_u*')
+
+-- Statusline string buffer
+local stlbuf = char_u_buf_t(256)
+local stlbuf_len = 256
+
+-- Expand statusline expression, returns a Lua string containing plaintext with only the characters
+-- that'll be displayed in the statusline
+function M.expand_statusline_expr(expr)
+    ffi.C.build_stl_str_hl(
+        ffi.C.curwin,
+        stlbuf,
+        stlbuf_len,
+        ffi.cast(char_u_str_t, expr),
+        0,
+        0,
+        0,
+        nil,
+        nil
+    )
+
+    return ffi.string(stlbuf)
+end
+
+-- Get display width of statusline expression
+function M.get_statusline_expr_width(expr)
+    return tonumber(ffi.C.build_stl_str_hl(
+        ffi.C.curwin,
+        stlbuf,
+        stlbuf_len,
+        ffi.cast(char_u_str_t, expr),
+        0,
+        0,
+        0,
+        nil,
+        nil
+    ))
+end
+
+return M
+
+
--- a/lua/feline/statusline_ffi.lua
+++ b/lua/feline/statusline_ffi.lua
@@ -4,7 +4,6 @@ local M = {}
 local ffi = require('ffi')
 
 -- Definitions required to use Neovim's build_stl_str_hl function to expand statusline expressions
--- And also other definitions used by this module
 ffi.cdef [[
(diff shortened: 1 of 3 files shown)


lua/feline/statusline_ffi.lua
```diff
@@ -0,0 +1,63 @@
...
+]]
+
+-- Used CType values stored in a local variable to avoid redefining them and improve performance
+local char8_buf_t = ffi.typeof('char8_t[?]')
+local char8_str_t = ffi.typeof('char8_t*')
+
+-- Expand statusline expression, returns a Lua string containing plaintext with only the characters
+-- that'll be displayed in the statusline
+function M.expand_statusline_expr(expr, winid)
+    local stlbuf = char8_buf_t(256)
```
shadmansaleh, line 38: You can create the buffer in module level and keep reusing that buffer for better performence .

lua/feline/statusline_ffi.lua
```diff
@@ -0,0 +1,63 @@
...
+    expr = ffi.cast(char8_str_t, expr)
+
+    ffi.C.build_stl_str_hl(
+        ffi.C.curwin,
```
shadmansaleh, line 42: Since you're expanding for winid you need to pass win_T pointer for that window. If you use curwin you'll get wrong results .

shadmansaleh, line 42: Since you're expanding for winid you need to pass win_T pointer for window of that winid else you'll get wrong results.

lua/feline/statusline_ffi.lua
shadmansaleh, line 38: You can create this buffer at module level and keep reusing it for better performence.

famiu: > You can create this buffer at module level and keep reusing it for better performance.

But then wouldn't the data be overriden everytime the function is used?

lua/feline/statusline_ffi.lua
famiu: > Since you're expanding for winid you need to pass win_T pointer for window of that winid else you'll get wrong results.

How would I go about doing that? Sorry, I'm really not familiar with Neovim's codebase

famiu: I figured it out. Though it'd be nice if I could reuse the window value to access its `w_width` member to avoid calling `api.nvim_win_get_width()`, but trying to do that makes Lua complain about the unknown size of the win_T struct.

famiu:
This seems to still have one issue, the inactive windows are not automatically truncated when a window is resized. This is more of a (Neo)Vim limitation since it doesn't have a `WinResized` autocommand

famiu:
It seems that using `WinScrolled` works to update the inactive windows

lua/feline/statusline_ffi.lua
shadmansaleh: You need to put `struct window_S` definition in `cdef` if you want to access `w_width` .

famiu: > You need to put `struct window_S` definition in `cdef` if you want to access `w_width` .

Yeah, I'm aware of that. But honestly that seems like more trouble than it's worth

shadmansaleh:
Instead of [setting local statusline option](https://github.com/famiu/feline.nvim/blob/a67897405f23cce54473d2afbc1a7df63688b815/lua/feline/init.lua#L40) for inactive windows you can use global option and draw inactive or inactive status based on which window you're in . I do that in lualine [just set an entry point in global option](https://github.com/shadmansaleh/lualine.nvim/blob/b7c0a2aeb1967e5b392429c04603d487e7f86bbe/lua/lualine.lua#L282) then [draw different things from that single entry point](https://github.com/shadmansaleh/lualine.nvim/blob/b7c0a2aeb1967e5b392429c04603d487e7f86bbe/lua/lualine.lua#L252-L277). That way inactive statuslines also gets dynamically updated just like active statusline.

famiu:
> Instead of [setting local statusline option](https://github.com/famiu/feline.nvim/blob/a67897405f23cce54473d2afbc1a7df63688b815/lua/feline/init.lua#L40) for inactive windows you can use global option and draw inactive or inactive status based on which window you're in . I do that in lualine [just set an entry point in global option](https://github.com/shadmansaleh/lualine.nvim/blob/b7c0a2aeb1967e5b392429c04603d487e7f86bbe/lua/lualine.lua#L282) then [draw different things from that single entry point](https://github.com/shadmansaleh/lualine.nvim/blob/b7c0a2aeb1967e5b392429c04603d487e7f86bbe/lua/lualine.lua#L252-L277). That way inactive statuslines also gets dynamically updated just like active statusline.
> 
> Or instead of triggering updates based on events you can update the inactive statuslines based on a timmer.

I had no idea g.actual_curwin even existed, that can actually make it possible to not pass winid to every function

shadmansaleh:
> I had no idea g.actual_curwin even existed, that can actually make it possible to not pass winid to every function

Read `:h stl` lol.

famiu:
> > I had no idea g.actual_curwin even existed, that can actually make it possible to not pass winid to every function
> 
> Read `:h stl` lol.

I did but I missed that part entirely somehow

famiu:
After rebasing it against the latest develop after #135. This seems to work really well and is also blazing fast, so I'm going to merge this now

PR #156, opened 2021-10-20
Add mode alias for "nt"

TheConfuZzledDude:
This function was broken by https://github.com/neovim/neovim/commit/f4359b5dbdd5ed9aa230532382edd7eb6cd0a61b which let `nvim_get_mode` return "nt" if in normal mode in term


famiu:
Change target branch to develop

PR #166, opened 2021-11-13
feat(vi_mode): Add nt mode

simrat39:
Used by termopen


famiu:
You should change the target branch to develop

Issue #164, opened 2021-11-04
[Feature] Use StatusLine highlight group by default


ldelossa:
**Is your feature request related to a problem? Please describe.**
I like this statusbar, but themeing it is a bit intense. 

**Describe the solution you'd like**
Would it be possible to set the default bg to the "StatusLine" highlight group so it "just works" with color schemes?

ldelossa:
I'm also noticing something like this is not possible:

```
local colors = { 
    bg = 'StatusLine',
    fg = "#e4e4e4"
}

require('feline').setup({
    colors = colors
})
```

This makes it difficult for individuals who theme switch,  as we expect StatusLine to change when the colorscheme changes.

vladimir-popov:
I think you can't use highlight group to define fg or bg color, because hg describes both of them. But, you can use a name of hg to specify highlight of your component:
```lua
{
     provider = "my_component",
     hl = "StatusLine"
}
```

famiu:
This has been done before but caused all sorts of other issues with highlight configuration such as the '^' character showing up in the inactive statusline due to Neovim's behavior caused by `StatusLine` and `StatusLineNC` having the same values, so it's not going to be done.

ldelossa:
@famiu fair enough. I've worked around this by just writing some lua anyway. 

One thing I'm still struggling with is, i often switch my vim theme from light to dark. With previous status lines using HL, the line colors would change as well. 

With Feline however, if i reload the module after the appropriate variables have changed, the status line just breaks. Is there a way to reload Feline during runtime? After its first setup

famiu:
> @famiu fair enough. I've worked around this by just writing some lua anyway.
> 
> One thing I'm still struggling with is, i often switch my vim theme from light to dark. With previous status lines using HL, the line colors would change as well.
> 
> With Feline however, if i reload the module after the appropriate variables have changed, the status line just breaks. Is there a way to reload Feline during runtime? After its first setup

What do you mean by "status line just breaks"? Are you talking about the highlights? If so, have you tried doing `require('feline').reset_highlights()`

ldelossa:
@famiu what i mean by breaks is, i attempt to reload the entire plugin with plenary's reload method, and when i do all elements on the status line disappear. 

I haven't tried the mentioned method. I will try this.

ldelossa:
@famiu so I attempted to call the function `require('feline').reset_highlights()` in my theme set function, and it results in a completely broken status line. 

[...]

PR #167, opened 2021-11-16
Feature: allows to specify a highlight group as a color

vladimir-popov:
This PR gives the ability to specify a color from existing highlight groups. It should help to create a status line that will use colors from the current `colorscheme`. 

For example, let's set 'Keyword' foreground color as the `fg` color in the theme of feline:

```lua
    use({
        'famiu/feline.nvim',
        config = function()
            local colors = require('feline.themes').default
            colors.fg = 'Keyword'
            require('feline').setup({
                theme = colors,
                components = require('feline.presets')['noicon']
            })
        end,
    })
```
Now, I have two schemes: `nightfox` Screenshot 2021-11-16 at 10 19 29 
and `onehalfdark` Screenshot 2021-11-16 at 10 19 54

So, the statusline will look different, depending on which scheme is used:

`nightfox`:  Screenshot 2021-11-16 at 10 23 42

`onehalfdark`: Screenshot 2021-11-16 at 10 23 23

related to #164

### DIFF
--- a/lua/feline/generator.lua
+++ b/lua/feline/generator.lua
@@ -60,6 +60,22 @@ local function add_hl(name, fg, bg, style)
     M.highlights[name] = true
 end
 
+-- Resolves color_or_hl depending on its content. 
+-- 
+-- @param color_or_hl String with or hex color '#xxxxxx', or name of highlight group.
+-- @param what        Which exactly color should be used in case of highlight group:
+--                    - 'foreground' to take a foreground color; 
+--                    - 'background' to take a background color.
+local function resolve_color(color_or_hl, what)
+    if (color_or_hl:sub(1, 1) == '#') or (vim.fn.hlID(color_or_hl) == 0)  then
+        return color_or_hl
+    end
+
+    local hl = vim.api.nvim_get_hl_by_name(color_or_hl, true)
+
+    return hl[what] and string.format('#%06x', hl[what]) or color_or_hl
+end
+
 -- Parse highlight table, inherit default/parent values if values are not given
 local function parse_hl(hl, parent_hl)
     parent_hl = parent_hl or {}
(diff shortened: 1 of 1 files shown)


famiu:
Why not make it more customizable and also allow setting background of one highlight group to foreground of a component and vice-versa

vladimir-popov:
@famiu this is a good idea, but then we should decide how exactly fg or bg should be specified. How about using the current logic in the case of a string with a name of hl,  and a table with key 'fg'/'bg' to specify particular color from hl:
```lua
-- a foreground color from the 'Keyword' will be taken: 
color.fg = 'Keyword' 

-- a background color from the 'Keyword' will be taken: 
color.fg = { 'Keyword' , 'bg' }
```

famiu:
> @famiu this is a good idea, but then we should decide how exactly fg or bg should be specified. How about using the current logic in the case of a string with a name of hl, and a table with key 'fg'/'bg' to specify particular color from hl:
> 
> ```lua
> -- a foreground color from the 'Keyword' will be taken: 
> color.fg = 'Keyword' 
> 
> -- a background color from the 'Keyword' will be taken: 
> color.fg = { 'Keyword' , 'bg' }
> ```

Honestly, this is partly why I feel like this PR is probably a bit unnecessary since `nvim_get_hl_by_name` already allows the user to get the fg or bg of a highlight and use it. Sure, this PR makes it a bit easier, but it's also incomplete if it only allows using the fg of the highlight group for the component fg and the bg of the highlight group for the component bg. Personally, I'm not a big fan of the syntax you suggested either, and I also can't think of a better way to frame it. If we wanted to do it the Feline way, then the way would be to add a utility function to get the fg / bg of a highlight, but as I said earlier, `nvim_get_hl_by_name` already allows that

vladimir-popov:
Ok, I get your point. So, PR can be closed. Nevertheless, thank you for the nice plugin!

PR #173, opened 2021-11-20
added option to turn off devicons on file.lua

FrenzyExists:
From https://github.com/famiu/feline.nvim/issues/172

Details are in the issue I made. TL;DR of the TL;DR, I added an option to disable devicons on file_info


famiu:
not sure why this is necessary since it's already possible to accomplish by setting the icon to an empty string

FrenzyExists:
Couldn't find a way to do such, and it was easier to me to just add on option to disable it right from the get-go

PR #176, opened 2021-12-02
refactor: use vim.diagnostic for lsp provider

ram02z:
Currently `diagnostic_warnings` and `diagnostic_info` don't work on 0.6 due to the severity name.

Since most of the functions now are `diagnostic` related, I wonder if changing the filename from `lsp.lua` will be more suitable.

Requires neovim 0.6, so probably hold off on merging or we could patch the two functions that don't work on neovim 0.6.

### DIFF
--- a/lua/feline/providers/lsp.lua
+++ b/lua/feline/providers/lsp.lua
@@ -8,7 +8,12 @@ function M.is_lsp_attached()
 end
 
 function M.get_diagnostics_count(severity)
-  return vim.tbl_count(diagnostic.get(0, { severity = severity }))
+    if vim.fn.has "nvim-0.6" then
+        return vim.tbl_count(diagnostic.get(0, { severity = severity }))
+    else
+        -- TODO: drop this when 0.5 is no longer used
+        return vim.lsp.diagnostic.get_count(0, ({ "Information", "Hint", "Warning", "Error" })[severity])
+    end
 end
 
 function M.diagnostics_exist(severity)
--- a/lua/feline/providers/lsp.lua
+++ b/lua/feline/providers/lsp.lua
@@ -1,23 +1,14 @@
 local M = {}
 
 local lsp = vim.lsp
+local diagnostic = vim.diagnostic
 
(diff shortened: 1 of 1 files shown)


kartikdc:
Please correct me if I am wrong, but on today's nightly update, nvim complains that vim.lsp.diagnostic.get_count() is deprecated so it would be a good time to merge this PR if it works with stable (0.6.*).

ram02z:
> Please correct me if I am wrong, but on today's nightly update, nvim complains that vim.lsp.diagnostic.get_count() is deprecated so it would be a good time to merge this PR if it works with stable (0.6.*).

Yeah this https://github.com/neovim/neovim/commit/c4d70dae802ef074aaf54bdcbbd5f73380f74a86 commit caused this. According to it, the warnings will stay until 0.7 where the old API functions will get removed. I think maybe it will be a good idea to merge into develop then, or even create a new version where the requirement changes from nvim 0.5 to 0.6, which will allow people on 0.5.x to pin it to v0.3.3.

zbirenbaum:
> > Please correct me if I am wrong, but on today's nightly update, nvim complains that vim.lsp.diagnostic.get_count() is deprecated so it would be a good time to merge this PR if it works with stable (0.6.*).
> 
> 
> 
> Yeah this https://github.com/neovim/neovim/commit/c4d70dae802ef074aaf54bdcbbd5f73380f74a86 commit caused this. According to it, the warnings will stay until 0.7 where the old API functions will get removed. I think maybe it will be a good idea to merge into develop then, or even create a new version where the requirement changes from nvim 0.5 to 0.6, which will allow people on 0.5.x to pin it to v0.3.3.
> 
> 

This is correct, but there's no need for new versions or anything. I made a similar PR to this one last night which switches the provider based upon whether the user has >=0.6 or now. Since it's necessary to keep nvim from complaining and doesn't break anything I considered it a hot fix and requested it go straight to master.

lukas-reineke (approved):
thank you :rocket:

PR #177, opened 2021-12-07
feat: add fileformat provider

mnacamura:
Just a small addition to show `fileformat` (`unix`, `dos`, or `mac`) in status line.


ram02z:
Thanks for the PR @mnacamura !

PR #179, opened 2021-12-13
update to vim.diagnostic, retain legacy functionality for older neovi…

### LINKED ISSUE #178: [Feature] Update to vim.diagnostic.get (if possible)
The import `feline.providers.lsp` causes the message
`vim.lsp.diagnostic.get_count is deprecated. See :h deprecated` to constantly popup in
the latest neovim nightly specifically under feline itself

Update to `vim.diagnostic.get` as `:h deprecated` suggests

Find some way to silence the warning as it will not go away (if possible)

**Additional context**
EDIT: It seems #176 has mentioned this already as well and provided a solution

zbirenbaum:
Fixes #178. Additionally, unlike #176, this change retains the legacy functionality. The table lookup is performed for legacy rather than new, since I think the current neovim major version should be supported by default, and extra calculations (no matter how fast or minor) should be reserved for legacy support.

The most recent neovim build complains about vim.lsp.diagnostic.get_count being deprecated, and it covers the command-line permanently, so this (or #176 if pre-0.6 functionality is unnecessary) really should be merged sooner rather than later. I made a PR for master instead of develop as I consider this a hotfix, given that it is really just code that prevents neovim from complaining.

### DIFF
--- a/lua/feline/providers/lsp.lua
+++ b/lua/feline/providers/lsp.lua
@@ -2,6 +2,13 @@ local M = {}
 
 local lsp = vim.lsp
 
+local severities = {
+  INFO="Information",
+  HINT="Hint",
+  WARN="Warning",
+  ERROR="Error",
+}
+
 function M.is_lsp_attached()
     return next(lsp.buf_get_clients(0)) ~= nil
 end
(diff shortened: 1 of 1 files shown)


lukas-reineke:
Thank you for this, but I liked the solution in #176 better.
I added support for 0.5 there

PR #181, opened 2021-12-19
v0.4.0

lukas-reineke:


lukas-reineke:
@famiu any reason to not merge this?
Is develop breaking anything?

famiu:
@lukas-reineke I think you need to rebase develop, your commits don't follow the conventional commit spec, also please don't use any merge commits, rebase changes instead

famiu:
I think we potentially *could* do with a v0.4.0 release now but I'm not exactly 100% sure if that's the best idea. In case we do choose to go with a release, please make sure to create a tag for the new release in the Git repo and make a [release note](https://github.com/feline-nvim/feline.nvim/releases) on GitHub for the release as well. You can see the previous release notes to use for reference.

lua/feline/providers/lsp.lua
```diff
@@ -3,16 +3,20 @@ local M = {}
...
+-- the diagnostic function to improve performance
+local severity_names = { "Information", "Hint", "Warning", "Error" }
+
 function M.is_lsp_attached()
     return next(lsp.buf_get_clients(0)) ~= nil
 end
 
 function M.get_diagnostics_count(severity)
-    if vim.fn.has "nvim-0.6" then
+    if vim.fn.has("nvim-0.6") == 0 then
```
ram02z, line 15: This condition is false for neovim >= 0.6. There is no need to check for 0.6

ram02z (changes requested):

lukas-reineke:
> I think we potentially could do with a v0.4.0 release now but I'm not exactly 100% sure if that's the best idea.

Your call. What would you like to change before releasing?

famiu:
> > I think we potentially could do with a v0.4.0 release now but I'm not exactly 100% sure if that's the best idea.
> 
> Your call. What would you like to change before releasing?

Nothing immediately comes to mind. Maybe try to add a fix for one of the open Issues that don't a big change to the codebase?

lukas-reineke:
IMO it is better to do many smaller releases, instead of collecting a bunch of stuff and then make a big release. (except when you bundle braking changes together)
I think if there is nothing super important, it is better to just release this and make a new release later. Especially because feline is currently really unusable with Neovim 0.6 if you use the LSP provider.

famiu:
> IMO it is better to do many smaller releases, instead of collecting a bunch of stuff and then make a big release. (except when you bundle braking changes together) I think if there is nothing super important, it is better to just release this and make a new release later. Especially because feline is currently really unusable with Neovim 0.6 if you use the LSP provider.

That makes sense. Go on with it, then. But don't forget to make a tag and a release and add release notes on GitHub

lukas-reineke:
I just saw that with [`f88bd9e` (#181)](https://github.com/feline-nvim/feline.nvim/pull/181/commits/f88bd9e77d5f1955d46417796fcbfdb45a364011) we break compatibility with 0.5 anyways.
Should I remove the compatibility check for the LSP provider? :thinking:

famiu:
> I just saw that with [`f88bd9e` (#181)](https://github.com/feline-nvim/feline.nvim/pull/181/commits/f88bd9e77d5f1955d46417796fcbfdb45a364011) we break compatibility with 0.5 anyways. Should I remove the compatibility check for the LSP provider? thinking

It's still compatible with 0.5, just that the truncation feature doesn't work without `nvim_eval_statusline`

PR #195, opened 2021-12-26
fix: only check buffer local gitsigns info

### LINKED ISSUE #194: [Bug] git branch shown even if not inside git repo
**Neovim version**
NVIM v0.6.0
Build type: Release
LuaJIT 2.0.5

**Feline version**
29f5cbf

**Describe the bug**
Git branch is shown even if not inside git repo. It's even shown in empty buffers and in vim help!
This happens after opening a file belonging to a git repo

**To Reproduce**
1. Open a file in a git repo
2. Branch is shown correctly
3. Open another file NOT in a git repo
4. The branch is shown (but it shouldn't)

**Expected behavior**
Branch should disappear. This is caused by this code:
https://github.com/feline-nvim/feline.nvim/blob/29f5cbfcecf563df84ac3b0e31e54e7b23e5331f/lua/feline/providers/git.lua#L33-L36
Why does it check `vim.g.gitsigns_head`? Feline should check only current buffer to detect correctly if it belongs to a git repo.

And should we keep `vim.g.gitsigns_head` in the provider?
https://github.com/feline-nvim/feline.nvim/blob/29f5cbfcecf563df84ac3b0e31e54e7b23e5331f/lua/feline/providers/git.lua#L6-L8

So I think we should remove `vim.g.gitsigns_head` from these files

ram02z:
When autochdir is disabled and user switches to a non-git file, the branch icon remains.

Closes #194

### DIFF
--- a/lua/feline/providers/git.lua
+++ b/lua/feline/providers/git.lua
@@ -1,5 +1,4 @@
 local b = vim.b
-local g = vim.g
 
 local M = {}
 
--- a/lua/feline/providers/git.lua
+++ b/lua/feline/providers/git.lua
@@ -4,7 +4,7 @@ local g = vim.g
 local M = {}
 
 function M.git_branch()
-    return b.gitsigns_head or g.gitsigns_head or '', ' '
+    return b.gitsigns_head or '', ' '
 end
 
 -- Common function used by the git providers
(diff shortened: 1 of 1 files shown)


xeluxee:
@ram02z it works well :)

I think we can remove the following line
https://github.com/feline-nvim/feline.nvim/blob/13de90cbefc863a89a81a313e5bf0e28490e6853/lua/feline/providers/git.lua#L2

ram02z:
> @ram02z it works well :)
> 
> I think we can remove the following line
> 
> https://github.com/feline-nvim/feline.nvim/blob/13de90cbefc863a89a81a313e5bf0e28490e6853/lua/feline/providers/git.lua#L2

Thanks

PR #198, opened 2021-12-28
Yank master -> develop

lukas-reineke:


lukas-reineke:
@famiu can you turn on merge commits?
Otherwise, we can't merge this without rebasing.

We have `Require linear history` on for develop, so it will never make a merge commit. For any branch, even with `Create a merge commit` it will always fast forward.

famiu:
> @famiu can you turn on merge commits? Otherwise, we can't merge this without rebasing.
> 
> We have `Require linear history` on for develop, so it will never make a merge commit. For any branch, even with `Create a merge commit` it will always fast forward.

Merge commits already seem to be turned on

lukas-reineke:
hm something in the branch settings then maybe :thinking: but I don't know which option would cause this

for now I'll just merge it locally

lukas-reineke:
THB i am not sure why it doesn't work...

i have `Require linear history` no for all my repos for the develop branch. And I can merge back master fine.

Maybe it is because we rebased develop against master before. So master is the base and not develop :thinking: 
But it shouldn't make a difference.

Issue #170, opened 2021-11-19
[Feature] Missing a utility function to get feedback whether a section is truncated or not


Tummetott:
Hello, 

to start with: Thanks for this amazing plugin, i recently switched from galaxyline because of the amazing truncation system. You just define priorities and and feline does the rest, love it! But now I'll try to describe my problem.

Part of my statusline looks like this:
![grafik](https://user-images.githubusercontent.com/19708025/142608670-11dfd114-7647-47c5-af80-c0e6109990e5.png)

The section with the `file_info` provider is configured with `truncate_hide = true`. When I decrease my window size, this section gets truncated and it looks like this

![grafik](https://user-images.githubusercontent.com/19708025/142608948-fe153719-2d4a-41e1-aaa1-6051c2c056aa.png)

The Problem is the highlighting of the previous `right_rounded` separator (I circled it with red). I would like to highlight the `bg` of the `right_rounded` separator with grey, when the `file_info` section is not truncated. But I would like to highlight it with the color of my terminal background, when the `file_info` section is truncated. However, I don't find a good way to get feedback from feline whether a section is truncated of not. 

I looked in the code of `generator.lua` and found e.g. the table `local component_widths = {}` which hold all the lengths of the sections during statusline generation. When a section is truncated, the respective length is 0. I know it's not a good idea to expose this specific table to the user, but a utility function like `require('feline').is_truncated(section)` would solve this issue. The function argument `section` could be a simple index since sections are not named. 

I don't feel confident enougth to make a pull request and hope, that someone can help me with this.

Best Tummetott

famiu:
I think a way better and more elegant solution would be to allow modifying the separators after truncation as well, like having a `short_separator` value, so you could change the highlight of that. Thoughts?

Tummetott:
In this case, a `short_separator` would not help because highlighting of the separator depends on the truncation state of the **next** section. Of course you could think of moving the separator in question to the next section but than I still have the same problem, when the separator highlighting depends on the truncation state of the next **two** sections. Think of my example with the screenshots and imagine I want to put another provider in the grey bar after the `file_info` provider.

famiu:
[...]

PR #218, opened 2022-02-13
feat: add utility functions to check component truncation

famiu:
Ref: #170

Adds two utility functions, `require('feline').is_component_truncated` and `require('feline').is_component_hidden` to check if a component has been truncated or hidden.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -164,6 +164,18 @@ end
 
 If you omit the provider value, it will be set to an empty string. A component with no provider or an empty provider may be useful for things like [applying a highlight to section gaps](#highlight-section-gaps) or just having an icon or separator as a component.
 
+#### Component name
+
+A component can optionally be given a name. While the component is not required to have a name and the name is mostly useless, it can be used to check if the component has been [truncated](#truncation). To give a component a name, just set its `name` value to a `string`, shown below:
+
+```
+local my_component = {
+    name = 'a_unique_name'
+}
+```
+
+Two components inside the `active` or `inactive` table cannot share the same name, so make sure to give all components unique names.
+
 #### Truncation
 
 Feline has an automatic smart truncation system where components can be automatically truncated if the statusline doesn't fit within the window. It can be useful if you want to make better use of screen space. It also allows you to better manage which providers are truncated, how they are truncated and in which order they are truncated.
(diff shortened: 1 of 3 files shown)


Tummetott:
Amazing 😻. Thank you so much !!

Tummetott:
Hmm don't know why but somehow it does not work. Consider this minimal example:

```lua
local left_section = {}
local right_section = {}

table.insert(left_section, {
    name = 'fancyname',
    provider = 'looooooooooooooooooooooooooong',
    short_provider = 'short',
})

table.insert(right_section, {
    provider = function ()
        if require('feline').is_component_truncated(vim.api.nvim_get_current_win(), 'fancyname') then
            return 'truncated'
        else
            return 'not truncated'
        end
    end,
})

local components = {
    active = {
        left_section,
        right_section
    },
}

require('feline').setup({
    components = components,
})
```

When I squeeze the window until the short provider is shown, the other provider is not affected. 
However when I run `lua print(require('feline').is_component_truncated(vim.api.nvim_get_current_win(), 'fancyname'))` as vim command it returns the right boolean value

Tummetott:
I see. Would it at least be possible to use the `is_component_truncated` and `is_component_hidden` functions inside hightlight functions of a component? Highlights don't affect the actual length of the component. This would be the actual use case why I opened issue #170 
However right now this is not possible too.

famiu:
> I see. Would it at least be possible to use the `is_component_truncated` and `is_component_hidden` functions inside hightlight functions of a component? Highlights don't affect the actual length of the component. This would be the actual use case why I opened issue #170 However right now this is not possible too.

but highlights (separator highlights, that is) depend on the provider output, and component highlight is calculated alongside the provider so it wouldn't be wise to move it somewhere else

Tummetott:
Hmm all right, I see the problem. So i guess there is no alternative left but to purge the last changes from the master branch.

famiu:
> Hmm all right, I see the problem. So i guess there is no alternative left but to purge the last changes from the master branch.

Why purge the changes? The utility function can still be used outside the components if ever necessary. Meanwhile, I'm trying to think of a way to make it work inside components

Tummetott:
Yeah u are right. I can think of use cases where the functions could be useful as they are. And I love you attitude!! Would like to sera workaround for this problem. But I don’t exactly know what you mean with `post_truncation` key

famiu:
> Yeah u are right. I can think of use cases where the functions could be useful as they are. And I love you attitude!! Would like to see a workaround for this problem. But I don’t exactly know what you mean with `post_truncation` key

so basically, it'd be a table inside the components table containing component values that need to be computed after truncation, like, with your example, it'd be:

```lua
local left_section = {}
local right_section = {}

table.insert(left_section, {
    name = 'fancyname',
    provider = 'looooooooooooooooooooooooooong',
    short_provider = 'short',
})

table.insert(right_section, {
    provider = 'text',
    post_truncation = {
        hl = {
            fg = function ()
                if require('feline').is_component_truncated(vim.api.nvim_get_current_win(), 'fancyname' then
                    return 'oceanblue'
                else
                    return 'purple'
                end
            end
        }
    }
})

local components = {
    active = {
        left_section,
        right_section
    },
}

require('feline').setup({
    components = components,
})
```

Of course that creates complications of its own, since the `provider` or a separator being changed means truncation has to be done again, which might change the values. But changing highlights of the component or separator shouldn't change anything. So maybe we could instead add a value to highlights to allow them to be evaluated after truncation?

famiu:
We could discuss this further in #224 if you want

PR #243, opened 2022-03-20
fix(providers/file_info): don't transform the empty filename

fsouza:
When opening an empty buffer and using the file_info provider with
opts.type = 'relative', the filename ends up becoming CWD. The solution
is to check that the filename is not empty, set it to '[No Name]' and
then skip all transformations.


famiu:
Change target branch to `develop`

Issue #249, opened 2022-04-04
[Bug] `file_info` provider doesn't display the right icon when buffer isn't linked to a file


xeluxee:
**Neovim version**
NVIM v0.6.1
Build type: Release
LuaJIT 2.1.0-beta3

**Feline version**
Latest 290bea8e8a782586de32bb8d0bfb056ff8acdb9c

**Describe the bug**
I've used to set an icon for special buffers, buffers not associated with a file, like NvimTree buffer, Outline buffer, Trouble buffer...
To set an icon for certain buffers you can use this snippet:
``` lua
require("nvim-web-devicons").set_icon {
  NvimTree = {
    icon = " ",
    color = "#46b7e1",
    name = "NvimTree"
  },
}

```
But when I open NvimTree buffer I see normal file icon instead of the icon I've set. Previously it used to work but recently the icon is displayed wrongly.

**Expected behavior**
Display the icon set with `require("nvim-web-devicons").set_icon`

**Possible solution**
When the following line
https://github.com/feline-nvim/feline.nvim/blob/290bea8e8a782586de32bb8d0bfb056ff8acdb9c/lua/feline/providers/file.lua#L66
is replaced with
``` lua
    local extension = vim.bo.filetype
```
everything seems to work fine.

famiu:
Fixed through #251

famiu:
I reverted #251 since that's not the correct way to get extensions. I've decided I'm not going to support filetype icons for buffers not attached to a file by default, but I assume it shouldn't be too hard to make a custom provider that does that.

xeluxee:
Before merging let us take some time to discuss, this issue is a bit complex
The change introduced with #251 works better for `CMakeLists.txt`, because it displays the gear icon (the correct one) instead of the common txt file icon
You can take as a comparison icons displayed in NvimTree
There are many different use cases and I think the design system of `nvim-web-devicons` should be improved because many plugins shows different icons on the same file due to api usage confusion (for example with the same `.bashrc` different icons are displayed in telescope, NvimTree and Feline :confounded:)

famiu:
[...]

PR #255, opened 2022-04-21
fix: use nvim-web-devicons correctly in file providers

xeluxee:
This should solve #249 

### How has this been tested?
With the following files/buffers:
- python and rust files (because extension != filetype)
- `CMakeLists.txt` (previously a wrong icon was displayed)
- `.bashrc` (previously a wrong icon was displayed)
- buffers not linked to a file (NvimTree, Outline, Trouble...)
- common files (where extension = filetype)

It seems to work fine for these use-cases and for the most common ones, but we'd need some time make sure edge-cases works well too

### DIFF
--- a/lua/feline/providers/file.lua
+++ b/lua/feline/providers/file.lua
@@ -62,16 +62,13 @@ local function get_unique_filename(filename, shorten)
 end
 
 function M.file_info(component, opts)
-    local filename = api.nvim_buf_get_name(0)
-    local extension = bo.filetype
-    local type = opts.type or 'base-only'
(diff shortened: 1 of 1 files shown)


famiu:
It seems your PR is based on `master` and not `develop`, you should fix that.
Also, `vim.fn.bufname()` returns the same thing as `vim.api.nvim_buf_get_name(0)`, so I see no reason behind replacing it. The only real change needed is sending `nil` as the extension and letting `nvim-web-devicons` deal with it.

xeluxee:
Closed in favour of #256

PR #261, opened 2022-05-06
feat: add `reverse` option to scroll bar provider

### LINKED ISSUE #258: [Feature] reverse scrollbar
no

desired: scrollbar provider reversed. should start with small sliver at top, grows towards bottom
current: starts with small sliver at bottom, grows towards top

n/a

**Additional context**
just seems more intuitive that the scrollbar provider's "animation" moves in the same direction as the scrolling through the file (i.e. downward)

famiu:
Closes #258

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -653,7 +653,7 @@ Feline by default has some built-in providers to make your life easy. They are:
 | [`vi_mode`](#vi-mode)                 | Current vim mode                               |
 | [`position`](#position)               | Get line and column number of cursor           |
 | `line_percentage`                     | Current line percentage                        |
-| `scroll_bar`                          | Scroll bar that shows file progress            |
+| [`scroll_bar`](#scroll-bar)           | Scroll bar that shows file progress            |
 | [`file_info`](#file-info)             | Get file icon, name and modified status        |
 | `file_size`                           | Get file size                                  |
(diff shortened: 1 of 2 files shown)


stumash:
so how do I actually supply the `{ reverse = true }` table as the `opts` argument here?

like what does a config look like that takes advantage of this? I don't want to have to manually write my own `components` table.

right now, I do something like

```lua
local default_components = require'feline.presets'.default
local my_components = {
  active = default_components.active,
  inactive = default_components.inactive
}
modify_my_components(my_components)
require'feline'.setup { components = my_components }
```

what do I need to do to get this reverse scrollbar going?

famiu:
> so how do I actually supply the `{ reverse = true }` table as the `opts` argument here?
> 
> like what does a config look like that takes advantage of this? I don't want to have to manually write my own `components` table.
> 
> right now, I do something like
> 
> ```lua
> local default_components = require'feline.presets'.default
> local my_components = {
>   active = default_components.active,
>   inactive = default_components.inactive
> }
> modify_my_components(my_components)
> require'feline'.setup { components = my_components }
> ```
> 
> what do I need to do to get this reverse scrollbar going?

```lua
provider = {
    name = 'scroll_bar'
    opts = {
        reverse = true
    }
}
```

PR #264, opened 2022-05-08
Fix nvim-web-devicons configured with highlight groups

### LINKED ISSUE #263: [Bug] Feline `file` provider not resolving the correct icon color
**Neovim version**
```
NVIM v0.7.0                                                                                                                    Build type: Release                                                                                                            LuaJIT 2.1.0-beta3                                                                                                             Compiled by builduser 
```

**Feline version**
`master` branch

**Describe the bug**
Feline `file` provider colored icon not respecting custom `nvim-web-devicons` highlight groups

**To Reproduce**
1. Set up custom highlight for `nvim-web-devicons`
2. Add `file` provider with colored icon
3. Check color

**Expected behavior**
I would expect the colors to match what I have set in the `nvim-web-devicons`. I am making sure that we are loading and setting up `nvim-web-devicons` before `feline` as well. Also I checked the code for `bufferline` (which works correctly) and they seem to be using the same `get_icon` function as well so I'm not sure what's causing the issue (https://github.com/akinsho/bufferline.nvim/blob/2d5266d14b7b5dbc54923469b4599f0349f7cedc/lua/bufferline/utils.lua#L262)

*Note:* It does seem to correctly resolve icons that I have added in my `nvim-web-devicon` setup correctly like if I add an icon for `.deb` or `.zip` that don't come out of the box but they show up not colored at all because it's not getting the correct color.

Here is a screenshot showing the resolved file type icon color in both bufferline (above and correct) and feline (below and incorrectly showing the default  color)

2022-05-08_15:12:06_screenshot

mehalter:
`nvim-web-devicons` sets up highlight groups for each icon if they don't exist so that it has more general compatibility with colorschemes adjusting the colors of icons. This adds support for this approach to theming `nvim-web-devicons`. We can guarantee these highlight groups exist because if they don't exist in the colorscheme that is loaded then `nvim-web-devicons` sets it up.

Also this is my first contribution so let me know if I need to change anything when it comes to formatting or structure. Thank you!

Resolves #263

### DIFF
--- a/lua/feline/providers/file.lua
+++ b/lua/feline/providers/file.lua
@@ -4,6 +4,11 @@ local api = vim.api
 
 local M = {}
 
+-- Get foreground color of highlight group (nvim-webdevicons guarantees these are set)
+local function get_hlgroup_fg(group)
+    return string.format('#%06x', vim.api.nvim_get_hl_by_name(group, true)['foreground'])
+end
+
 -- Get the names of all current listed buffers
 local function get_current_filenames()
     local listed_buffers = vim.tbl_filter(function(bufnr)
@@ -66,7 +71,7 @@ function M.file_info(component, opts)
 
     -- Avoid loading nvim-web-devicons if an icon is provided already
     if not component.icon then
-        local icon_str, icon_color = require('nvim-web-devicons').get_icon_color(
+        local icon_str, hlgroup = require('nvim-web-devicons').get_icon(
             fn.expand('%:t'),
             nil, -- extension is already computed by nvim-web-devicons
             { default = true }
@@ -75,7 +80,7 @@ function M.file_info(component, opts)
         icon = { str = icon_str }
 
         if opts.colored_icon == nil or opts.colored_icon then
-            icon.hl = { fg = icon_color }
+            icon.hl = { fg = get_hlgroup_fg(hlgroup) }
         end
     end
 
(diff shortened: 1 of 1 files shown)


mehalter:
It looks like this basically reverts #111 going through the pull requests in the past. I think that this is a much better approach for supporting much better compatibility with themes.

mehalter:
This might be considered a bug upstream actually. I made a commend on `nvim-web-devicons` to ask about it. If they agree to fix this upstream then this PR is probably not the best approach to this.

famiu:
> This might be considered a bug upstream actually. I made a commend on `nvim-web-devicons` to ask about it. If they agree to fix this upstream then this PR is probably not the best approach to this.

Let's wait for their statement then, I personally think this would be better fixed upstream

mehalter:
Yeah I agree @famiu I'll keep pushing on that repo and maybe I'll just open a PR to fix this functionality to help in the process as well. I'll post an update here when I hear back! Thanks for your consideration in the meantime

mehalter:
I went ahead and opened a PR to `nvim-web-devicons` to fix this upstream: https://github.com/kyazdani42/nvim-web-devicons/pull/128

🤞🏼

thieung:
@famiu @mehalter 
They've just merged the fix 🚀

famiu:
@mehalter You can go on and close this Pull Request and also #263 if the problem is fixed

mehalter:
Thanks for letting me know @famiu ! Gonna go ahead and close this. Sorry for the initial noise on this repo for upstream bug

PR #275, opened 2022-05-27
fix: escape special statusline chars in filename component

dsych:
`jdt.ls` generates filenames with `%` in them e.g. 
```
jdt://contents/java.base/java.lang/StringBuilder.class?=jacoco/%5C/usr%5C/lib%5C/jvm%5C/java-17-openjdk-17.0.3.0.7-1.fc36.x86_64%5C/lib%5C/jrt-fs.jar%60java.base=/maven.pomderived=/true=/=/ja
vadoc_location=/https:%5C/%5C/docs.oracle.com%5C/en%5C/java%5C/javase%5C/17%5C/docs%5C/api%5C/=/%3Cjava.lang(StringBuilder.class
```
The `file_info` component reverse searches for the first occurrence of `/`, which in this case will contain `%3Cjava.lang(StringBuilder.class`. Unfortunately, `%` is a special character in the statusline (as you probably know), which causes feline to crash (in my case it's the winbar that's crashing).

Something to consider: you might want extract this escaping logic into a user-facing utils function. For instance, I also display the alt file in my winbar, which would also benefit from the same sanitazation.

### DIFF
--- a/lua/feline/providers/file.lua
+++ b/lua/feline/providers/file.lua
@@ -120,6 +120,8 @@ function M.file_info(component, opts)
         modified_str = ''
     end
 
+    -- escape any special statusline characters in the filename
+    filename = filename:gsub('%%', '%%%%')
     return string.format('%s%s%s', readonly_str, filename, modified_str), icon
 end
 


famiu:
Replace the double quotes with single quotes

dsych:
The way i see this, it's more about semantics. i agree that function as simple as that does not make sense of it's own. however given fact that 99% of ppl are not going to look into it, but *will* appreciate the convenience of not having to think about which characters are allowed, i think it's still worth it.

Anyway, it's really up to you, i am not pushing it.

famiu:
> The way i see this, it's more about semantics. i agree that function as simple as that does not make sense of it's own. however given fact that 99% of ppl are not going to look into it, but _will_ appreciate the convenience of not having to think about which characters are allowed, i think it's still worth it.
> 
> Anyway, it's really up to you, i am not pushing it.

I'll consider it, let's merge this for now

PR #290, opened 2022-07-25
fix(providers): add missing vim modes

edwarmv:
Source ```h: mode()```


lua/feline/providers/vi_mode.lua
```diff
@@ -12,8 +12,11 @@ local mode_alias = {
     ['niR'] = 'NORMAL',
     ['niV'] = 'NORMAL',
     ['v'] = 'VISUAL',
+    ['vs'] = 'VISUAL',
     ['V'] = 'LINES',
+    ['Vs'] = 'LINES',
     [''] = 'BLOCK',
+    ['s'] = 'BLOCK',
```
famiu, line 19: `s` is used twice

edwarmv: Line 19 stands for ```CTRL-Vs```
from neovim help: mode()
```CTRL-Vs  Visual blockwise using |v_CTRL-O| in Select mode```

PR #298, opened 2022-09-01
feat: added "slashes" option to file_info provider

Nick-pr:
Added the ability for users to define a substitute for the slashes in a path. 

For example, 

`foo/bar/test.lua`

could turn into

`foo > bar > test.lua`

usage of this feature would look like this with these changes.

```lua
{
  provider = {
    name = "file_info",
    opts = { slashes = " > "},
  }
}
```

### DIFF
--- a/doc/feline.txt
+++ b/doc/feline.txt
@@ -970,8 +970,8 @@ through the provider `opts`:
     modified.<br> Default:`'●'`
 - `file_readonly_icon` (string): The icon that is shown when a file is
     read-only.<br> Default:`'🔒'`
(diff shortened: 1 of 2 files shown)


famiu:
I don't like the name `slashes`, I think something like `path_sep` is better

Nick-pr:
> I don't like the name `slashes`, I think something like `path_sep` is better. Also, your code probably doesn't work on Windows.

I agree I think `path_sep` is more clear. For now, I'll close this PR and open an issue instead because I don't have the ability to test properly in windows. Perhaps when I have more time.

PR #304, opened 2022-09-13
feat: add path_sep option to file_info provider

### LINKED ISSUE #301: [Feature] Change the path seperator for the file_info provider
As a user, I would like the ability to replace the path separators within a path when using the default provider `file_info`.

For example,
 this `foo/bar/test.lua`
 could be `foo > bar > test.lua`

Usage could look something like this when defining a component:
```lua
{
  provider = {
    name = "file_info",
    opts = { path_sep = " > "}  
  }
}
```

I think the simplest and most intuitive way of implementing this is just by adding an option for the `file_info` provider

**Considerations**
- Should work correctly on both Windows and UNIX machines.

ram02z:
`get_path_separator` function is originally from the plenary path module.
https://github.com/nvim-lua/plenary.nvim/blob/4b66054e75356ac0b909bbfee9c682e703f535c2/lua/plenary/path.lua#L21

Closes #301

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -708,7 +708,7 @@ The `file_info` provider has some special provider options that can be passed th
 - `file_readonly_icon` (string): The icon that is shown when a file is read-only.<br>
   Default:`'🔒'`
 - `path_sep` (string): The separator character in the file path. <br>
-  Default: `/` or `\\` depending on OS 
+  Default: `/` or `\` depending on OS 
 - `type` (string): Determines which parts of the filename are shown. Its value can be one of:
 
   - `'full-path'`: Full path of the file (eg: `'/home/user/.config/nvim/init.lua'`)
--- a/USAGE.md
+++ b/USAGE.md
@@ -707,6 +707,8 @@ The `file_info` provider has some special provider options that can be passed th
   Default:`'●'`
 - `file_readonly_icon` (string): The icon that is shown when a file is read-only.<br>
   Default:`'🔒'`
+- `path_sep` (string): The separator character in the file path. <br>
+  Default: `/` or `\\` depending on OS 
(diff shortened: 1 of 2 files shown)


famiu (approved):

famiu:
need to add docs, apart from that everything else seems fine at first glance

lua/feline/providers/file.lua
```diff
@@ -61,6 +61,22 @@ local function get_unique_filename(filename, shorten)
...
 end
 
+-- Get path separator depending on OS
+local function get_path_separator()
+    if jit then
+        local os = string.lower(jit.os)
+        if os == 'linux' or os == 'osx' or os == 'bsd' then
+            return '/'
+        else
+            return '\\'
```
famiu, line 71:
```suggestion
            return [[/]]
        else
            return [[\]]
```

Raw strings look cleaner in this case imo

USAGE.md
```diff
@@ -707,6 +707,8 @@ The `file_info` provider has some special provider options that can be passed th
   Default:`'●'`
 - `file_readonly_icon` (string): The icon that is shown when a file is read-only.<br>
   Default:`'🔒'`
+- `path_sep` (string): The separator character in the file path. <br>
+  Default: `/` or `\\` depending on OS
```
famiu, line 711:
```suggestion
  Default: `/` or `\` depending on OS 
```
People might confuse the `\\` for a literal `\\`. It's better to just include only the character without any escaping done to avoid amibuity. Lua will show an unescaped string error if they try to use a singular `\` without a raw string anyway.

famiu (approved):

USAGE.md
ram02z: Good point

PR #324, opened 2022-10-26
feat: Added `macro` provider

LucasTavaresA:
Apparently checking for nil will keep the indicator active, just like the hlsearch one


LucasTavaresA:
oof i forgot stylua

famiu (approved):

PR #339, opened 2023-01-23
Remove unnecessary spacing before line percentage

SystematicError:
By specifying `%2d` in the `string.format`, when the line percentage is a single digit value, there is a unwanted space at the beginning of the string, with this PR, the functionality should continue to work exactly as intended whilst avoiding that extra padding before the string.

### DIFF
--- a/lua/feline/providers/cursor.lua
+++ b/lua/feline/providers/cursor.lua
@@ -58,7 +58,7 @@ function M.line_percentage()
     elseif curr_line == lines then
         return 'Bot'
     else
-        return string.format('%2d%%%%', math.ceil(curr_line / lines * 99))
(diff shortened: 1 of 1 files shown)


famiu:
This behavior is intended to prevent the statusline components from unnecessarily moving around. An extra space doesn't really hurt anybody either.

SystematicError:
Thanks for the input, the space was a bit weird for me since I was grouping components together and at times the component was spaced a bit too much, should I look into making this into a togglable option?

famiu:
> Thanks for the input, the space was a bit weird for me since I was grouping components together and at times the component was spaced a bit too much, should I look into making this into a togglable option?

I guess we could add a `padding` opt to `line_percentage`

Issue #319, opened 2022-10-04
[Bug] Actually disable winbar when using `disable` option


FollieHiyuki:
**Neovim version**
```
NVIM v0.8.0
Build type: MinSizeRel
LuaJIT 2.1.0-beta3
Compiled by buildozer@build-edge-x86_64

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```

**Feline version**
`master` branch

**Describe the bug**
Unlike statusline, when adding filetypes to `disable.filetypes` list (e.g. NvimTree) I expect the winbar to not be there at all instead of an empty winbar.

**To Reproduce**
```lua
require('feline').winbar.setup {
    components = {
        active = {
             {
                 { provider = "Hello" },
             },
        },
    },
    disable = {
        filetypes = { 'NvimTree' },
    }
}
```

**Expected behavior**
The winbar is completely gone in the defined buffer.

**Does this error occur in the minimal init file?**

**Provide modified minimal_init.lua**

**Screenshots**
There is a blank line on the top (the winbar) of NvimTree buffer.

![2022-10-04-224214_grim](https://user-images.githubusercontent.com/67634026/193864496-b4467502-3b48-4c99-adae-e9359514f0cd.png)

**Additional context**
Looking briefly at the code it is returning an empty string. I think it should return `nil` in the winbar case.

https://github.com/feline-nvim/feline.nvim/blob/5d6a054c476f2c2e3de72022d8f59764e53946ee/lua/feline/generator.lua#L533-L535

famiu:
This would not work as it is not allowed by Neovim to disable winbar on a per-window basis.

FollieHiyuki:
Thanks for answering!

Would it be ok for feline.nvim to support setting winbar through `autocmd` with window option vim.wo.winbar (instead of setting it globally)? This way I can achieve the same result.

famiu:
> Thanks for answering!
> 
> Would it be ok for feline.nvim to support setting winbar through `autocmd` with window option vim.wo.winbar (instead of setting it globally)? This way I can achieve the same result.

That would be an antipattern. Plus there's already plans to add support for selectively disabling winbar for Neovim, so let's wait for that

PR #340, opened 2023-02-14
feat: add winbar autocmd option

Ursescu:
-> Allow autocmd winbar option as a workaround in order to actually disable winbar on a per-window basis.

Indeed there was an issues some time ago: #319. I agree that autocmd is an antipattern, but until we have native support, I think it's useful to have this as a temporary solution. It's very annoying to have winbar on NvimTree and qf windows.


Ursescu:
After some testing I concluded that this is not working properly.

